PR #3, opened 2021-10-12
Wrap Google Storage API

mveytsman:
* FakeMediaStorage for testing which records the last call made to it
* LocalMediaStorage for development which writes to the local file
  system
* GoogleMediaStorage for production/development which uses the GCP
  API

Mirror of https://github.com/bikebrigade/core/pull/124 @drteeth 

- [x] Dev is happy
- [x] Test is happy
- [x] Staging works


mveytsman:
@drteeth Staging instructions here: https://github.com/bikebrigade/dispatch#staging

it's then available at staging.bikebrigade.ca - we use an old copy of the database but your login should work.

PR #4, opened 2021-10-12
Update show.html.heex

dkagotho:
Added target blank for the delivery map buttons and created an open in google map [button[](url)](url)

### DIFF
--- a/lib/bike_brigade_web/live/delivery_live/show.html.heex
+++ b/lib/bike_brigade_web/live/delivery_live/show.html.heex
@@ -250,7 +250,7 @@
       <h2 class="ml-7 text-base font-semibold text-gray-900">
         Full Route Map
       </h2>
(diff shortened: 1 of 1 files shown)


mveytsman (changes requested):
I think this looks a little off with the button being below the header.


Take a look at the code for the I'm done button

```html
<div class="flex flex-row items-center justify-between">
          <h2 class="mr-1 text-base font-semibold text-gray-900">
            Tell us you're done.
          </h2>
          <div class="flex flex-wrap items-start justify-end flex-shrink ml-auto">
            <C.button href={sms_done_url(@campaign, @rider)} class="ml-auto"
              color={:green}>
              <%= Heroicons.Outline.chat_alt(class: "w-5 h-5 mr-1 -ml-0.5") %>
                I'm done!
            </C.button>
          </div>
        </div>
```
You want a `flex flex-row items-center justify-between` on the div that contains the header and the button, similar to how we do it above

PR #5, opened 2021-10-13
Login feature test

drteeth:
* Adds [Wallaby](https://hexdocs.pm/wallaby/readme.html) for Feature/Integration testing
* Adds an example test for Login

- [x] Works locally
- [x] Works in CI


mveytsman:
@drteeth have you tried https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html at all?

I think that it can do everything Wallaby can for liveviews, and is faster / doesn't need the chrome dependency.

The only thing that it *wouldn't* cover is the test you wrote here which is a "deadview" but that's because when I wrote the login page LiveView didn't have hooks yet - it can be rewritten as a liveview.

Feels on trying some LiveView tests to see if you like them?

mveytsman:
So browser tests more for like full integration testing, and liveview tests for "browser unit tests?"

lib/bike_brigade_web/endpoint.ex
```diff
@@ -1,6 +1,10 @@
 defmodule BikeBrigadeWeb.Endpoint do
   use Phoenix.Endpoint, otp_app: :bike_brigade
 
+  if sandbox = Application.get_env(:bike_brigade, :sandbox) do
```
mveytsman, line 4: @drteeth can you explain how this works? I'm curious why we didn't have it before (i guess we didnt exercise the endpoint in the LV tests?)

mveytsman (approved):
LGTM

Looks like the sandbox stuff lets us make our tests async 🚀

PR #18, opened 2021-11-01
Disable scheduled messaging in dev

mveytsman:
@dkagotho this disables the scheuled messenger that caused all the log lines when you were in the dev console

@drteeth how do you think this looks / compared with the adapter stuff which is related but different?


mveytsman:
via @drteeth and @jennaleeb maybe I should just disable logging here

PR #2, opened 2021-10-12
Duplicate campaign

mveytsman:
Mirror of https://github.com/bikebrigade/core/pull/110

cc @jennaleeb

PR #22, opened 2021-11-04
Copy riders and tasks

jennaleeb:
Follow up to https://github.com/bikebrigade/dispatch/pull/2

Add ability to optionally copy over riders and delivery tasks

1. Worth pushing to staging to poke at it. 
2. Should the checkboxes be checked by default?
3. Will make some comments inline about things I think might be weird

### DIFF
--- a/lib/bike_brigade/delivery/task.ex
+++ b/lib/bike_brigade/delivery/task.ex
@@ -18,6 +18,44 @@ defmodule BikeBrigade.Delivery.Task do
     removed: "removed"
   )
 
+  @fields [
+    :contact_email,
+    :contact_name,
+    :contact_phone,
+    :delivery_date,
+    :delivery_window,
+    :dropoff_address,
+    :dropoff_address2,
+    :dropoff_city,
+    :dropoff_email,
+    :dropoff_location,
+    :dropoff_name,
+    :dropoff_organization,
+    :dropoff_phone,
+    :dropoff_postal,
+    :dropoff_province,
+    :delivery_status,
+    :delivery_status_notes,
+    :onfleet_dropoff_id,
+    :onfleet_pickup_id,
+    :organization_name,
+    :organization_partner,
+    :other_items,
+    :pickup_address,
+    :pickup_address2,
+    :pickup_city,
+    :pickup_country,
+    :pickup_location,
+    :pickup_postal,
+    :pickup_province,
+    :request_type,
+    :rider_notes,
+    :size,
+    :submitted_on,
+    :assigned_rider_id,
+    :campaign_id
+  ]
+
   # TODO we're missing a dropoff_country field :joy:
   schema "tasks" do
     field :contact_email, :string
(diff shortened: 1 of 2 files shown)


lib/bike_brigade/delivery/task.ex
```diff
@@ -257,4 +259,10 @@ defmodule BikeBrigade.Delivery.Task do
         changeset
     end
   end
+
+  def fields_for(task) do
```
jennaleeb, line 263: I hate this method name (for `campaign` as well). Anything better? "Normalized fields" or something?

lib/bike_brigade/delivery/task.ex
```diff
@@ -257,4 +259,10 @@ defmodule BikeBrigade.Delivery.Task do
...
+    fields = @fields
+    |> Enum.filter(fn field -> field not in [:submitted_on, :assigned_rider_id, :campaign_id] end)
```
jennaleeb, line 265: Not sure if there is a better way to do this

lib/bike_brigade_web/live/campaign_live/duplicate_campaign_component.ex
```diff
@@ -57,26 +82,40 @@ defmodule BikeBrigadeWeb.CampaignLive.DuplicateCampaignComponent do
...
 
-    {:ok, new_campaign} = create_new_campaign(old_campaign, duplicate_form)
+    {:ok, new_campaign} = create_new_campaign(old_campaign, date_time_form_params)
 
     if old_campaign.instructions_template_id do
       copy_instructions_template(old_campaign, new_campaign)
     end
 
+    if duplicate_deliveries == "true" do
+      copy_delivery_tasks(old_campaign, new_campaign)
```
jennaleeb, line 101: Probably okay to leave as a follow up but if there are a lot of tasks/riders this could be really slow & time out. Is there a phoenix/elixer-y way of handling this in a background job?

lib/bike_brigade_web/live/campaign_live/duplicate_campaign_component.ex
```diff
@@ -100,4 +139,37 @@ defmodule BikeBrigadeWeb.CampaignLive.DuplicateCampaignComponent do
...
+      task_params = Delivery.Task.fields_for(old_task)
+      {:ok, new_task} = Delivery.create_task_for_campaign(new_campaign, task_params)
+
+      old_task = old_task |> Repo.preload(:task_items)
+
+      for task_item <- old_task.task_items do
+        %Delivery.TaskItem{}
+        |> Delivery.TaskItem.changeset(%{task_id: new_task.id, item_id: task_item.item_id, count: task_item.count})
+        |> Repo.insert()
+      end
```
mveytsman, line 154:
```suggestion
      old_task = old_task |> Repo.preload(:task_items)
      task_params = Delivery.Task.fields_for(old_task)
        |> Map.put(:task_items,  for task_item <- old_task.task_items, do: %{ item_id: task_item.item_id, count: task_item.count})
      {:ok, new_task} = Delivery.create_task_for_campaign(new_campaign, task_params)
      end
```

mveytsman (approved):
👍  had a suggestion to make the copy tasks a little cleaner - can you confirm it works right?

lib/bike_brigade_web/live/campaign_live/duplicate_campaign_component.ex
jennaleeb: Hmm it wont compile, and I'm getting

```
** (SyntaxError) lib/bike_brigade_web/live/campaign_live/duplicate_campaign_component.ex:147: unexpected comma. Parentheses are required to solve ambiguity in nested calls.

This error happens when you have nested function calls without parentheses. For example:

    one a, two b, c, d

In the example above, we don't know if the parameters "c" and "d" apply to the function "one" or "two". You can solve this by explicitly adding parentheses:

    one a, two(b, c, d)

Or by adding commas (in case a nested call is not intended):

    one, a, two, b, c, d

Elixir cannot compile otherwise. Syntax error before: ','
```

Might just be syntax error, but not sure where to fix?

jennaleeb:
@mveytsman I'm going to merge as is, feel free to take a look at the error message when you have a chance and we can refactor!

PR #27, opened 2021-11-09
move shell.nix to niv

ghedamat:


mveytsman (approved):
Thanks @ghedamat! This works for me -- and it solves the problem I was doing in a much worse way when I was trying to pin nixpkgs.

PR #30, opened 2021-11-10
Add more tests

mveytsman:
- adding riders in campaigns#show


mveytsman:
@dkagotho I think this is good to merge

PR #33, opened 2021-11-13
Rider show page version .9

chadmohr:
This is the initial HTML for a new rider page. It will likely need some edits and updates and changes once populated with live data.

image


jennaleeb:
Hey pals! I made a few updates:

1. Keep the ability to click a rider's name to see their location on the map (without immediately navigating away)
2. Update the `message` button to navigate to the rider's conversation
3. Populate the rider's campaign info with some data (haven't touched the campaign history/schedule stuff yet, and also can likely be a lot cleaner 💅)

lib/bike_brigade_web/live/rider_live/show.ex
```diff
@@ -0,0 +1,71 @@
...
+  defp deliveries(rider_stats) do
+    rider_stats |> Enum.map(& &1.task_id) |> Enum.uniq() |> Enum.count
+  end
+
+  defp distance(rider_stats) do
+    rider_stats |> Enum.map(& &1.distance) |> Enum.sum()
+  end
+
+  defp latest_campaign_info(rider_stats) do
+    latest_campaign_id = rider_stats |> Enum.map(& &1.campaign_id) |> Enum.max(fn -> nil end)
```
jennaleeb, line 61: Using the max ID is a bit icky, maybe something like?

```
campaign_ids = rider_stats |> Enum.map(& &1.campaign_id)
query = Ecto.Query.from(c in Campaign, where: c.id in ^campaign_ids, order_by: [desc: c.delivery_start])
```

Though the whole `rider_stats` query is maybe a bit weird too.

lib/bike_brigade_web/live/rider_live/show.ex
```diff
@@ -0,0 +1,71 @@
...
+    query = from r in Rider,
+      where: r.id == ^rider.id,
+      join: t in assoc(r, :assigned_tasks),
+      join: c in assoc(t, :campaign),
+      select: %{rider_id: r.id, campaign_id: c.id, task_id: t.id, distance: t.delivery_distance}
+
+    Repo.all(query)
+  end
+
+  defp campaigns(rider_stats) do
```
mveytsman, line 48: we can probably do these in SQL as 

```elixir
query = from r in Rider,
      where: r.id == ^rider.id,
      join: t in assoc(r, :assigned_tasks),
      join: c in assoc(t, :campaign),
      select: %{campaigns: count(c.id, :distinct), tasks: count(t.id, :distinct), distance: sum(t.delivery_distance)}
```

lib/bike_brigade_web/live/rider_live/show.ex
mveytsman: On second thought check out https://github.com/bikebrigade/dispatch/pull/36/files

PR #35, opened 2021-11-20
Scripts improvements

mveytsman:
MacOS uses a different path for default chrome -- we now support both

I pulled the wrapping in nixos into a function.


drteeth (approved):
LGTM.

PR #36, opened 2021-12-01
Add view-backed ecto schemas for Rider stats

drteeth:


mveytsman:
@drteeth @jennaleeb and I decided to add a check for the start time being <= NOW() and merge this. We think that we should still cound faile deliveries for your stats, e.g. if a rider failed to contact a person and still delivered to a community fridge it counts!

I also moved the SQL directly into migrations from the files. My thinking is that I do want versioning for the views that migrations provide that having them in files wouldn't achieve.

drteeth:
I was thinking that for views in migrations you have to copy the body of the migration forward in a new migration when you want to update it. You get versioning, but you lose the diff. If you have separate view files, you still have to make a new, largely empty, migration every time but you keep the history on the view.

It's not great either way, and I'd rather do it on the fly but I couldn't think of a better way to have it integrate with Ecto. I wish we could back a schema on a query, ecto or sql.

PR #38, opened 2021-12-07
Add tags to rider

jennaleeb:


jennaleeb:
It's tags! @mveytsman maybe we can try it out on staging before merging?

jennaleeb:
oops nvm 1 thing still missing (creating a new tag). stand by!

jennaleeb:
I've added https://github.com/bikebrigade/dispatch/pull/38/commits/8beccf52c36ec2d9dc0cf62d10f99e790be52d15 to create the tag if it doesn't appear in the dropdown. It means removing the ability to use "Enter" to submit the form, but I think that is okay? Lmk if there is another UX pattern that would make more sense.

PR #40, opened 2021-12-08
Create UI.table

mveytsman:
This PR rewrites most of the tables to use `UI.table` which is a component that abstracts out most of the css for tables

What's missing
- the printable tables: they use some css tricks so may as well keep these seperate
- opportunities: there's a component for editing the row inline,

### DIFF
--- a/lib/bike_brigade_web/live/program_live/form_component.html.heex
+++ b/lib/bike_brigade_web/live/program_live/form_component.html.heex
@@ -81,45 +81,36 @@
       <%= label f, :default_item_id, class: "block text-sm font-medium leading-5 text-gray-700" do %>
         Items
       <% end %>
-      <table class="min-w-full table-fixed">
-        <thead>
-          <tr>
-            <th class="w-1/3 px-6 py-3 text-xs font-medium leading-4 tracking-wider text-left text-gray-500 uppercase border-b border-gray-200 bg-gray-50">
-              Name
-            </th>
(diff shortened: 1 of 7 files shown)


lib/bike_brigade_web/live/program_live/form_component.html.heex
```diff
@@ -81,45 +81,36 @@
...
+        <:th class="w-1/3">
+          Name
+        </:th>
+        <:th class="w-full">
+          Description
+        </:th>
+        <:th class="w-1/3">
+          Category
+        </:th>
+        <:th class="flex justify-end w-32" padding="px-3 py-3" uppercase={false}>
```
jennaleeb, line 94: curious why `uppercase` and `padding` are variables vs the other extra "custom" classes that seem to just get appended?

jennaleeb (approved):

lib/bike_brigade_web/live/program_live/form_component.html.heex
mveytsman: The implementation might make it clearer: https://github.com/bikebrigade/dispatch/pull/40/files#diff-7792a950d4b26c637bba8528adf397d20b1eab976a175e9922b382fe1328df7eR209

tailwind classes just get appended as you said, so if we want to change padding and put it in the `class` field we'd end up with extra `p-X`'s. Same with uppercase, I want it default for all headers except when you say no.

I don't think this is the clearest way to do it but I couldn't find a better one :(

PR #49, opened 2022-01-06
Allow a different pg username to be set in .env file

jennaleeb:
I hit the `ERROR 42501 (insufficient_privilege)` whenever I forget to rename `postgres` => `jennablumenthal` (aka all the time). This makes my workflow slightly easier (and I don't think should break anything else?).


mveytsman (approved):
Thanks for doing this!

PR #50, opened 2022-01-07
When location fails geocoding, fail gracefully and post error message to slack

jennaleeb:
Closes https://trello.com/c/AeMXGEsS/88-dont-fail-importing-list-of-riders-from-mailchimp-if-location-fails-geocoding

### DIFF
--- a/lib/bike_brigade/importers/mailchimp_importer.ex
+++ b/lib/bike_brigade/importers/mailchimp_importer.ex
@@ -7,6 +7,7 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
   alias BikeBrigade.Importers.Importer
   alias BikeBrigade.Riders
   alias BikeBrigade.Riders.Rider
+  alias BikeBrigade.Messaging.SlackWebhook
 
   @count 100
   @mailchimp "mailchimp"
@@ -136,15 +137,20 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
       max_distance = translate_max_distance(member.merge_fields[:RADIO16])
       capacity = translate_capacity(member.merge_fields[:RADIO17])
 
-      {:ok, location} =
-        Location.geocoding_changeset(%Location{}, %{
-          address: address,
-          postal: postal,
-          city: city,
-          province: province,
-          country: country
-        })
-        |> Ecto.Changeset.apply_action(:save)
+      location_changeset = Location.geocoding_changeset(%Location{}, %{
+        address: address,
+        postal: postal,
+        city: city,
+        province: province,
+        country: country
+      })
+      location = case Ecto.Changeset.apply_action(location_changeset, :save) do
+          {:ok, location} -> location
+          {:error, _errors} ->
+            error_message = "Address error for ${:name} (${:email}): ${:address}. Please look up their rider page and edit manually!"
+            Task.start(SlackWebhook, :post_message, [error_message])
+            %Location{address: "1 Front St"}
+      end
 
       rider_attrs = %{
         mailchimp_id: member.id,
--- a/lib/bike_brigade/importers/mailchimp_importer.ex
+++ b/lib/bike_brigade/importers/mailchimp_importer.ex
@@ -1,19 +1,16 @@
 defmodule BikeBrigade.Importers.MailchimpImporter do
   import Ecto.Query, warn: false
-  alias Ecto.Multi
   import BikeBrigade.Utils, only: [get_config: 1, with_default: 2]
   alias BikeBrigade.Repo
   alias BikeBrigade.Location
   alias BikeBrigade.Importers.Importer
   alias BikeBrigade.Riders
-  alias BikeBrigade.Riders.Rider
   alias BikeBrigade.Messaging.SlackWebhook
 
   @count 100
   @mailchimp "mailchimp"
   @import_failed_tag %{name: "Dispatch Import Failed", status: "active"}
 
-  # TODO make this a singleton? / add locking
   def sync_riders(opts \\ nil) do
     Repo.transaction(fn ->
       last_synced =
@@ -30,8 +27,10 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
         case parse_mailchimp_attrs(member) do
           {:ok, rider_attrs} ->
             create_or_update(rider_attrs)
+
           {:error, {:update_location, rider_attrs}} ->
-            {:ok, rider} = rider_attrs
+            {:ok, rider} =
+              rider_attrs
               |> put_default_location()
               |> create_or_update()
 
@@ -41,6 +40,12 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
             notify_import_error(member, error)
         end
       end
+
+      Repo.insert(%Importer{name: @mailchimp, data: %{last_synced: DateTime.utc_now()}},
+        returning: true,
+        on_conflict: update_data_map(),
+        conflict_target: :name
+      )
     end)
   end
 
@@ -70,14 +75,16 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
     error_message = """
       We had trouble with the address for #{rider.name}. Please edit manually: https://dispatch.bikebrigade.ca/riders/#{rider.id}/edit
     """
+
     Task.start(SlackWebhook, :post_message, [error_message])
   end
 
   def notify_import_error(member, error) do
-    error_message = "An error ocurred when importing #{member.email_address} #{Kernel.inspect(error)}"
+    error_message =
+      "An error ocurred when importing #{member.email_address} #{Kernel.inspect(error)}"
+
     Task.start(SlackWebhook, :post_message, [error_message])
   end
-
 
   def get_members(last_changed \\ nil) do
     with {:ok, account} <- Mailchimp.Account.get(),
@@ -154,8 +161,11 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
       }
 
       case geolocate_raw_location(raw_location) do
-        {:ok, location} -> {:ok, Map.put(rider_attrs, :location_struct, Map.from_struct(location))}
-        {:error, _} -> {:error, {:update_location, rider_attrs}}
+        {:ok, location} ->
+          {:ok, Map.put(rider_attrs, :location_struct, Map.from_struct(location))}
+
+        {:error, _} ->
+          {:error, {:update_location, rider_attrs}}
       end
     else
       {:error, err} ->
@@ -181,7 +191,8 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
       province: location.province,
       country: location.country
     })
-    |> Ecto.Changeset.apply_action(:save) #TODO: this should be a method on the location struct to validate
+    # TODO: this should be a method on the location struct to validate
+    |> Ecto.Changeset.apply_action(:save)
   end
 
   # TODO: do we need this?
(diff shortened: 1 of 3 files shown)


lib/bike_brigade/messaging/slack_webhook.ex
```diff
@@ -40,6 +40,23 @@ defmodule BikeBrigade.Messaging.SlackWebhook do
     :ok = SlackApi.send_webook(payload)
   end
 
+  def post_message(message) do
```
mveytsman, line 43: I think we're still needing to manage two webhook URLs in the config. Let me know if you want a 🍐 on this

jennaleeb: Ya I haven't touched this since we chatted on tuesday. I want to also attempt the refactor so that the rider is created and we have the `id` to pass to slack.

lib/bike_brigade/importers/mailchimp_importer.ex
```diff
@@ -13,78 +13,71 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
   @mailchimp "mailchimp"
   @import_failed_tag %{name: "Dispatch Import Failed", status: "active"}
 
-  def sync_existing_rider(%Rider{} = rider) do
```
jennaleeb, line 15: Dead code, we can always come back to this PR if we need it?

test/bike_brigade/importers/mailchimp_importer_test.exs
```diff
@@ -0,0 +1,50 @@
...
+    status: "subscribed",
+    timestamp_opt: "2019-08-24T14:15:22Z",
+    merge_fields: %{
+      FNAME: "Morty",
+      PHONEYUI_: "647-555-5555",
+      PHONE: "647-555-5555",
+      TEXTYUI_3: "1508 Davenport Rd"
+    }
+  }
+  test "parse_mailchimp_attrs is successful if fields from Mailchimp are valid" do
```
jennaleeb, line 18: `parse_mailchimp_attrs` should probably be untested and a private method.. was keeping it around for debugging and because `sync_riders` handles the errors without returning anything. thoughts welcome!

lib/bike_brigade/importers/mailchimp_importer.ex
```diff
@@ -7,83 +7,77 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
...
-         {:ok, rider_attrs} <- build_rider(member) do
-      Riders.update_rider(rider, rider_attrs)
-    end
-  end
-
-  def sync_riders do
-    Multi.new()
-    |> Multi.run(:get_last_synced, fn repo, _changes ->
+  # TODO make this a singleton? / add locking
+  def sync_riders(opts \\ nil) do
```
jennaleeb, line 17: Added a trello card to properly mock Mailchimp without needing this https://trello.com/c/hL9bC1HX/93-create-a-fakemailchimp-to-allow-for-better-mocking-without-api-credentials

lib/bike_brigade/messaging/slack_webhook.ex
jennaleeb: I am going to tackle the slack refactor in a separate PR. Discussed earlier (leaving here for posterity), we are currently using the [`Incoming Webhooks`](https://api.slack.com/messaging/webhooks) API. This has a disadvantage in that each app installed in each channel has its own (private) URL. Meaning we'd have an `.env` file that looks something like

```
channel_a="some-secret-url"
channel_b="another-secret-url"
channel_c="yet-another-secret-url"
```

If we use [`chat.postMessage`](https://api.slack.com/methods/chat.postMessage), we can use a single auth token stored in the `.env` file, and commit the `channel_ids` to the repo as they are not sensitive.

lib/bike_brigade/importers/mailchimp_importer.ex
```diff
@@ -7,83 +7,77 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
...
         end
 
-      {:ok, last_synced}
-    end)
-    |> Multi.run(:get_mailchimp_members, fn _repo, %{get_last_synced: last_synced} ->
-      get_members(last_synced)
-    end)
-    |> Multi.merge(fn %{get_mailchimp_members: mailchimp_members} ->
-      multi = Multi.new()
+      {:ok, members} = opts || get_members(last_synced)
```
mveytsman, line 26: I think it's weird that we're passing `{:ok, [...]}` in tests

you can pass in the list of members itself, or better yet create an inner function that takes a list of members 

```elixir
def sync_riders() do
     Repo.transaction(fn -> 
        last_synced = ... # see comment above
        {:ok, members} = get_members(last_synced) # evaluate whether we need to return an `{:ok, members}` here - we needed the :ok tuples in the multi, but now we probably want to throw / revert the transaction if mailchimp is down
        process_members(members)
end

def process_members(members) do
      Repo.transaction(fn -> 
         ....
      end)
end
```

If you're asking yourself about the nested transactions, calling `Repo.transaction` inside another transaction is a no-op which is what we want -- see last paragraph https://hexdocs.pm/ecto/Ecto.Repo.html#c:transaction/2-use-with-function

jennaleeb: I'm not sure I'm following? We want to skip the `get_members` function in tests because it will get angry about a missing `api_key`. I feel like the better solution is to create a `FakeMailchimp` module (similar to `FakeSlack`). Thoughts? Maybe I am missing how you write the tests with the inner function.

lib/bike_brigade/importers/mailchimp_importer.ex
```diff
@@ -7,83 +7,77 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
...
+  # TODO make this a singleton? / add locking
+  def sync_riders(opts \\ nil) do
+    Repo.transaction(fn ->
       last_synced =
-        case repo.get_by(Importer, name: @mailchimp) do
+        # TODO add a row lock here.
+        case Repo.get_by(Importer, name: @mailchimp) do
           nil -> nil
           importer -> importer.data["last_synced"]
         end
```
mveytsman, line 24:
```suggestion
 Repo.one(
          from i in Importer,
            where: i.name == ^"mailchimp",
            lock: "FOR UPDATE SKIP LOCKED",
            select: fragment("? ->> 'last_synced'", i.data)
        )
```

This is what the row lock looks like - it will cause all other transactions to get back nil while the transaction is running. we are locking this particular row of the importer tablw, which corresponds to just the mailchimp importer.

A good refactor would be to put that in the Importers module but I think this is the only one that needs it so far so fine to wait.

lib/bike_brigade/importers/mailchimp_importer.ex
mveytsman: Oh sure, and have it be configurable to return some members.

I'm saying you could also just test the inner function

test/bike_brigade/importers/mailchimp_importer_test.exs
```diff
@@ -15,36 +22,78 @@ defmodule BikeBrigade.Importers.MailchimpImporterTest do
...
+  describe "sync_riders/0" do
+    setup do
+      saved_env = Application.get_env(:bike_brigade, BikeBrigade.Importers.MailchimpImporter)
 
-    {:ok, _} = MailchimpImporter.sync_riders({:ok, [invalid_attrs]})
-  end
+      Application.put_env(
+        :bike_brigade,
+        BikeBrigade.Importers.MailchimpImporter,
+        Keyword.put(saved_env, :list_id, @list_id)
```
jennaleeb, line 53: Interesting! Is this updating some kind of global state? Any links to elixir testing info I can read more about?

lib/bike_brigade/importers/mailchimp_importer.ex
```diff
@@ -77,9 +78,20 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
     Map.put(rider_attrs, :location_struct, %{address: "1 Front St"})
   end
 
+  def tag_invalid_location(rider) do
+    # TODO make this an idempotent Riders.add_tag
+    rider =
+      rider
+      |> Repo.preload(:tags)
+
+    tags = Enum.map(rider.tags, & &1.name) ++ ["invalid_location"]
```
jennaleeb, line 87: Do you think people will remember to remove the tag once they fix the address?

test/bike_brigade/importers/mailchimp_importer_test.exs
```diff
@@ -103,5 +103,21 @@ defmodule BikeBrigade.Importers.MailchimpImporterTest do
...
+
+      FakeMailchimp.clear_members(@list_id)
+
+      FakeMailchimp.add_members(@list_id, [
+        put_in(@valid_attrs, [:merge_fields, :FNAME], "Mortimus")
+      ])
+
+      assert {:ok, _} = MailchimpImporter.sync_riders()
+      r = Riders.get_rider_by_email("dispatcher@example.com")
+      assert r.name == "Mortimus"
```
jennaleeb, line 120: 😛

lib/bike_brigade/importers/mailchimp_importer.ex
mveytsman: I hope so! I added a note to the slack message

test/bike_brigade/importers/mailchimp_importer_test.exs
mveytsman: Yup  -- the environment is a global key/value state https://hexdocs.pm/elixir/Application.html#module-the-application-environment

The stuff in config.exs and dev.exs/test.exs/prod.exs ends up being put there at compile time, and the stuff in runtime.exs gets there at runtime.

I could have put a list id in test.exs, but I felt that it didn't belong there since it was an arbitrary string used in mocking our importer and not an actual testing configuration thing.

The docs for ExUnit.Case are a good intro to testing: https://hexdocs.pm/ex_unit/1.12/ExUnit.Case.html (I generally find the elixir docs excellent but you need to know which module name you're looking for to get to the good stuff :( )

It might be useful to look at the changes I made in your tests: https://github.com/bikebrigade/dispatch/pull/50/files/96b3a0bda0498eb5d738369d27d90504c7ae1f10..4c38173f6a33492cf6ccac080862b5045de65a15#diff-15a82d00038d613b51a9716bfafd946bbe606c7e3b3018f3c287c2cc5cfc8de0

I think the main things I introduced were
1) `assert` works with pattern matching as well, e.g. `assert {:ok, _} = foo()` -- just putting the pattern match without the assert will still error and fail the test, but the assert will give the correct feedback and make it a test failure instead of an exception.
2) Added some `describe`s, this is pretty self explanatory but note you only have one level of `describe`s. Jose talks about some of the thinking on that (and being stubborn about adding it at all) on this excellent recent podcast episode: https://podcast.thinkingelixir.com/82
3) `setup/on_exit` to setup the application environment. See https://hexdocs.pm/ex_unit/1.12/ExUnit.Callbacks.html#on_exit/2 and that module's doc for the test lifecycle.

I didn't use them, but [test contexts](https://hexdocs.pm/ex_unit/1.12/ExUnit.Callbacks.html) are a little counterintuitive and used in a lot of other tests.

If you want to go deeper, note that the test starts with `use BikeBrigade.DataCase`. This is a module that lives in our project - it's been autogenerated for us by phoenix. Worth taking a look: https://github.com/bikebrigade/dispatch/blob/main/test/support/data_case.ex

If you want to go *really* deep, the [code for `assert`](https://github.com/elixir-lang/elixir/blob/a64d42f5d3cb6c32752af9d3312897e8cd5bb7ec/lib/ex_unit/lib/ex_unit/assertions.ex#L45) is a great example of the power of macros

PR #51, opened 2022-01-11
Emoji Keyboard

dkagotho:
Added an emoji button in messages. Working on adding an icon instead of the word Emoji.

Screen Shot 2022-01-11 at 6 45 15 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 2022-01-12
+  - Added emoji keyboard 
 # 2021-12-06
   - Add location picker to campaigns
 # 2021-11-02
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -347,13 +347,23 @@ let liveSocket = new LiveSocket("/live", Socket, {
 Hooks.EmojiButtonHook = {
   mounted() {
     const picker = new EmojiButton();
-    const trigger = document.querySelector("#emoji-button");
-    picker.on("emoji", selection => {
-      document.querySelector("#conversation-form_body").value += selection.emoji;
-  })
-    trigger.addEventListener("click", () => picker.togglePicker(trigger));
-  }
+    const inputEl = document.querySelector(`#${this.el.dataset.inputId}`);
+    if (inputEl === undefined) {
+      console.warn("Input element for EmojiButton ", this.el.id, " not found. Emoji button disabled.")
+      return;
+    }
 
+    picker.on("emoji", selection => {
+       inputEl.value += selection.emoji;
+    });
+
+    picker.on("hidden", () => {
+       const end = inputEl.value.length;
+       inputEl.setSelectionRange(end, end);
+       inputEl.focus();
+    });
+    this.el.addEventListener("click", () => picker.togglePicker(this.el));
+  }
 }
 
 // Show progress bar on live navigation & form submits w/ delay of 100 ms
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -17,7 +17,7 @@ import "@ryangjchandler/alpine-clipboard"
 import "alpinejs"
 import Tribute from "tributejs";
 import Chart from 'chart.js/auto';
-
+import { EmojiButton } from '@joeattardi/emoji-button';
 
 
 let Hooks = {}
@@ -343,6 +343,18 @@ let liveSocket = new LiveSocket("/live", Socket, {
     }
   }
 })
+
+Hooks.EmojiButtonHook = {
+  mounted() {
+    const picker = new EmojiButton();
+    const trigger = document.querySelector("#emoji-button");
+    picker.on("emoji", selection => {
+      document.querySelector("#conversation-form_body").value += selection.emoji;
+  })
+    trigger.addEventListener("click", () => picker.togglePicker(trigger));
+  }
+
+}
 
 // Show progress bar on live navigation & form submits w/ delay of 100 ms
 let progressTimeout
--- a/lib/bike_brigade_web/live/campaign_live/messaging_form_component.html.heex
+++ b/lib/bike_brigade_web/live/campaign_live/messaging_form_component.html.heex
@@ -22,14 +22,17 @@
     autocomplete="off" phx-submit="save">
   <div class="flex flex-row mt-2 space-x-3">
     <div class="w-1/2">
-      <%= label f, :body, class: "block text-sm font-medium leading-5 text-gray-700" do %>
-        Text Message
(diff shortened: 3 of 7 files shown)


assets/js/app.js
```diff
@@ -344,6 +344,18 @@ let liveSocket = new LiveSocket("/live", Socket, {
   }
 })
 
+Hooks.EmojiButtonHook = {
+  mounted() {
+    const picker = new EmojiButton();
+    const trigger = document.querySelector("#emoji-button");
```
mveytsman, line 350: Our next step is to abstract the id of the element out of there. We can have multiple emoji buttons on one page (imagine a page with several text areas that have emoji keyboards). We need a way for the `mounted()` callback to know the element it corresponds to, and to communicate the input element we want to be adding the emoji to from the backend.

dkagotho:
Screen Shot 2022-01-12 at 6 33 36 PM
Screen Shot 2022-01-12 at 6 33 56 PM

lib/bike_brigade_web/live/campaign_live/messaging_form_component.html.heex
```diff
@@ -22,14 +22,17 @@
     autocomplete="off" phx-submit="save">
   <div class="flex flex-row mt-2 space-x-3">
     <div class="w-1/2">
-      <%= label f, :body, class: "block text-sm font-medium leading-5 text-gray-700" do %>
-        Text Message
+      <%= label f, :text, class: "block text-sm font-medium leading-5 text-gray-700" do %>
+        Text Message 
+        <C.button id="emoji-button" phx-hook="EmojiButtonHook" data-input-id={input_id(f, :textbody)} type="button" color={:white} size={:xxsmall}>
```
mveytsman, line 27: IDs for html elements need to be unique. You have two different elements with the same ID `emoji-button`. This is causing some of your problems

lib/bike_brigade_web/live/campaign_live/messaging_form_component.html.heex
```diff
@@ -53,10 +56,13 @@
     <div class="w-1/2">
       <%= label f, :details, class: "block text-sm font-medium leading-5 text-gray-700" do %>
         App Details
+        <C.button id="emoji-button" phx-hook="EmojiButtonHook" data-input-id={input_id(f, :details_body)} type="button" color={:white} size={:xxsmall}>
```
mveytsman, line 59: I think you're missing what `input_id(f, :details_body)` does. 

This example from the docs may help https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#textarea/3

```
# Assuming form contains a User schema
textarea(form, :description)
#=> <textarea id="user_description" name="user[description]"></textarea>
```

Note that in this example we called `textarea` *without* and id, and it automatically got the id `user_description`

In our case we called the `textarea` function *with* an id parameter which overrides the default

```
<%= textarea instructions_template_form, :body, required: true,
            id: "text-body",
```

This means the resulting element was
```
<textarea id="text-body">
....
```

The `input_id` function https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#input_id/2 gives you the default id for an input field given it's name.

As I said in my previous comment, you could have written the button as
```
<C.button id="emoji-button" phx-hook="EmojiButtonHook" data-input-id="text-body" type="button" color={:white} size={:xxsmall}>
```

But we're using the `input_id` function which generates the id for us. 

Try getting rid of the id parameter entirely and seeing what the results are like in the inspector. See if you can get the code to work without hardcoding theid.

lib/bike_brigade_web/live/sms_message_live/form_component.html.heex
```diff
@@ -21,12 +21,16 @@
         Message
       <% end %>
       <div class="mt-1 rounded-md shadow-sm">
-        <%= textarea f, :body, required: true, phx_debounce: "blur", class: "block w-full px-3 py-2 placeholder-gray-400 transition duration-150 ease-in-out border border-gray-300 rounded-md appearance-none focus:outline-none focus:ring-blue focus:border-blue-300 sm:text-sm sm:leading-5" %>
+        <%= textarea f, :sms_message_form, required: true, phx_debounce: "blur", class: "block w-full px-3 py-2 placeholder-gray-400 transition duration-150 ease-in-out border border-gray-300 rounded-md appearance-none focus:outline-none focus:ring-blue focus:border-blue-300 sm:text-sm sm:leading-5" %>
```
mveytsman, line 24: this parameter is the name of the input field, that's what we use on the backend to determine what to do with that input. I think changing it will break the entire form.

lib/bike_brigade_web/live/sms_message_live/form_component.html.heex
```diff
@@ -21,12 +21,16 @@
...
       </div>
       <%= error_tag f, :body %>
     </div>
 
     <div class="flex justify-end mt-6">
+                                                                                                <%# This looks wrong %>
+      <C.button class="mr-2" id="emoji-button" phx-hook="EmojiButtonHook" data-input-id={input_id(f, :sms_message_form)} type="button" color={:white} size={:xxsmall}>
```
mveytsman, line 31: You have two buttons with the same id on the page (look in the inspector to confirm this!) this is the cause of some of your problems.

lib/bike_brigade_web/live/campaign_live/messaging_form_component.html.heex
```diff
@@ -53,10 +55,12 @@
     <div class="w-1/2">
       <%= label f, :details, class: "block text-sm font-medium leading-5 text-gray-700" do %>
         App Details
+        <C.button id="emoji-button" phx-hook="EmojiButtonHook" data-input-id={input_id(f, :details)} type="button" color={:white} size={:xxsmall}>
```
mveytsman, line 58: Let's make the ids different here too
```suggestion
        <C.button id="emoji-button2" phx-hook="EmojiButtonHook" data-input-id={input_id(f, :details)} type="button" color={:white} size={:xxsmall}>
```

mveytsman (approved):
😁 🎉

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+
+# 2022-01-12
+  - Added emoji keyboard 
 # 2022-01-13
   - Edit tags for riders
 # 2021-12-06
```
mveytsman, line 6:
```suggestion
# 2022-01-14
  - Added emoji keyboard 
# 2022-01-13
  - Edit tags for riders
# 2021-12-06
```
Updating date and making sure the order is correct

mveytsman (approved):

PR #52, opened 2022-01-12
Add script to restore staging db from production

mveytsman:
This requires the user to have two environment variables `PROD_DATABASE_URL` and `STAGING_DATABASE_URL`

Not sure how we will communicate these, perhaps 1password?

### DIFF
--- a/scripts/restore_staging_db
+++ b/scripts/restore_staging_db
@@ -13,17 +13,12 @@ if [ "$STAGING_DATABASE_URL" == "" ]; then
   exit 1
 fi
 
-DUMP_PATH="$(dirname "$0")/../tmp/db_dump"
-
-#!/bin/bash
-function finish {
-  echo "Cleaning up"
-  rm $DUMP_PATH
-}
-trap finish EXIT
-
-echo "Dumping production database"
-wrap_nix_shell "pg_dump $PROD_DATABASE_URL -Fc -f $DUMP_PATH"
-echo "Restoring to staging"
-wrap_nix_shell "pg_restore --clean --no-acl -d $STAGING_DATABASE_URL $DUMP_PATH"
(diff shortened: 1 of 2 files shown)


mveytsman:
This script caused a small incident where Chad was bombarded with messages from staging.

My message to slack below
---

I have written a script (https://github.com/bikebrigade/dispatch/pull/52) to copy the production database to staging so we can manually sync them, but there was two issues:
It wasn’t doing an entirely clean restore which would put the staging database in a weird state. @Jenna (she/her) ran into this testing something on staging.
We copied over the table that contains scheduled messages, so if there was an unsent scheduled message available at the time of copying the database, staging would also try to send it out.
The combination of 1 and 2 meant that staging kept trying to send scheduled messages, and couldn’t write to the database saying that the scheduled message was sent, causing a cascade of sending Chad the same message over and over again from the staging app. Luckily, on staging we only send SMSs to people with dispatcher accounts, so this would not have affected riders who aren’t also dispatchers. AFAIK it just affected Chad based on who was signed up to deliver at that particular time.
I’ve fixed 1 by shutting down the service and dropping the staging database entirely during this sync process, and 2 by explicitly not copying the data from the scheduled messages table. Both changes are in https://github.com/bikebrigade/dispatch/pull/52/commits/d5c68d755a230b1226b12583db1537ff8366e254
I’ll include this text in the PR.
@Ben Moss (he/him) do you have any thoughts on making this more resilient? I feel like we’ll end up with another table that shouldn’t be copied over… A long term solution that I want to head towards is having this database copying also modify the data to anonymize it on staging, where we can exclude the scheduled messages. Any other thoughts on making this whole process more resilient long or short term?
As I’m writing this, maybe the scheduled messages live in a special schema for queues which we can ignore?

PR #56, opened 2022-01-26
Message length

dkagotho:
Created a word counter that changes colour. When the words are below 1200 words, the colour is teal,
Screen Shot 2022-01-26 at 11 48 45 AM

When the words are above 1200 and are less than 1400, the colour changes to amber.
Screen Shot 2022-01-26 at 11 50 26 AM

When the words are above 1401 the colour changes to red.
Screen Shot 2022-01-26 at 11 51 36 AM


mveytsman (approved):
This looks fantastic thanks @dkagotho

PR #57, opened 2022-01-27
Refactor Slack Api

jennaleeb:
Previously, we have been using the [`Incoming Webhooks`](https://api.slack.com/messaging/webhooks) API. This has a disadvantage in that each app installed in each channel has its own (private) URL. Meaning we'd have an `.env` file that looks something like

```
channel_a="some-secret-url"
channel_b="another-secret-url"
channel_c="yet-another-secret-url"
```

In this PR, we are using [`chat.postMessage`](https://api.slack.com/methods/chat.postMessage), with a single auth token stored in the `.env` file, and so commit the `channel_ids` to the repo as they are not sensitive.

~❗️Before this merges:~ (done)

 We need to choose which app we already have installed (or install a new one) and add that to our `.env` file (in fly.io?)


lib/bike_brigade/messaging/slack.ex
```diff
@@ -0,0 +1,21 @@
...
+    @channel_id "C01QU0YVACW" #dispatch
+
+    def post_message(message) do
+      payload = SlackApi.PayloadBuilder.build(@channel_id, message)
+      :ok = SlackApi.post_message(payload)
+    end
+  end
+
+  defmodule Operations do
+    @channel_id "C016VGHETS4" #software
```
jennaleeb, line 14: Should we set these in the config so we can have a different channel for dev/prod?

mveytsman: I like that!

You can just do something like 

```
config :bike_brigade, BikeBrigade.Messaging.Slack.Operations,
   chanel_id: "C016VGHETS4"
   channel_name: "software"
```

lib/bike_brigade/slack_api.ex
```diff
@@ -7,7 +7,13 @@ defmodule BikeBrigade.SlackApi do
 
   @callback post!(url, body, headers) :: :ok
 
-  @headers [{"content-type", "application/json"}]
+  @token BikeBrigade.Utils.fetch_env!(:slack, :token)
```
mveytsman, line 10: this may not work in production -- the module attribute `@token` is set at compile time, and compilation happens in an environment without the secrets which are loaded at runtime.

I think you need to put the token in the headers at runtime

lib/bike_brigade/messaging/slack.ex
```diff
@@ -0,0 +1,21 @@
+defmodule BikeBrigade.Messaging.Slack do
+  alias BikeBrigade.SlackApi
+
+  defmodule RiderSms do
+    @channel_id "C01QU0YVACW" #dispatch
+
+    def post_message(message) do
```
mveytsman, line 7: I think there's two or three different things that are being complected by `RiderSms` and `Operations` both having the same function with the same argument, and both calling `SlackApi.PayloadBuilder.build` - which decides what the slack message will look like based on the type of the input.

For example, `Operations.post_message(%SmsMessage{...}` will end up posting a message formatted as a rider sms message to #software, and `RiderSms.post_message("hello")` will post "hello" to the dispatch channel.

The naming of these modules makes me thing they are about differentiating the *kind* of slack message we're sending (with also an associated channel). I think it makes sense to move the payload construction to its own module, but maybe the payload builder shouldn't just change based on type?

jennaleeb: So, it was entirely intentional to _not_ differential the kind of slack messages we send based on the channel name (since we'd maybe want to send different messages with different formatting requirements to the same channel). Here we just use the modules as little helpers to set the correct message destination (other option is doing something like `Slack.message("body", "channel-name")` which is also okay but feels a bit magic-string/atom-y).

Agreed there is a bit of weirdness between the type of message/destination, though I think it's partially because we only have the two use cases at the moment. Thoughts about leaving it as is and returning to it once we have a 3rd use case that might help inform the design?

config/prod.exs
```diff
@@ -6,6 +6,14 @@ config :bike_brigade, BikeBrigadeWeb.Endpoint,
   cache_static_manifest: "priv/static/cache_manifest.json",
   force_ssl: [rewrite_on: [:x_forwarded_proto], hsts: true]
 
+config :bike_brigade, BikeBrigade.Messaging.Slack.Operations,
```
mveytsman, line 9: I think we should have staging use the api-playground channel as well - can you move this into `runtime.exs`, there's an if statement that decides whether its staging or production 

If its unclear the config stuff is kind of dodgy - prod.exs is when the compile-time environment is production which is true for staging and production. runtime.exs can decide what the runtime env is, which is set with an environment variable (see https://github.com/bikebrigade/dispatch/blob/main/fly.toml#L6-L7 vs https://github.com/bikebrigade/dispatch/blob/main/fly.staging.toml#L6-L7 for how its set)

lib/bike_brigade/importers/mailchimp_importer.ex
```diff
@@ -94,14 +94,14 @@ defmodule BikeBrigade.Importers.MailchimpImporter do
       We had trouble with the address for #{rider.name}. Please edit manually: #{BikeBrigadeWeb.Router.Helpers.rider_index_url(BikeBrigadeWeb.Endpoint, :edit, rider.id)}, and don't forget to remove the `invalid_location` tag!
     """
 
-    Task.start(SlackWebhook, :post_message, [error_message])
+    Task.start(Slack.Operations, :post_message, [error_message])
```
mveytsman, line 97: I think that `:post_message` should be renamed to `:post_message!` in order to indicate that it can raise an exception (it calls `post!` which has the bang)

My thinking is that if we used `Task.start_link` here instead of `Task.start` the exception would crash the import.

mveytsman (commented):
This looks great! I had two small suggestions before we :ship:

lib/bike_brigade/messaging/slack.ex
mveytsman: One more thought  - this would be a good place to look into [behaviours](https://elixir-lang.org/getting-started/typespecs-and-behaviours.html#behaviours)

config/prod.exs
jennaleeb: Gotcha! Ya I was a bit confused between `runtime`/`config`! Will move.

mveytsman (approved):
👍

PR #62, opened 2022-02-08
Update show.html.heex

chadmohr:
A variety of CSS fixes to rider page for better display. Includes new markup for rider details, fixed height with internal scrolling for schedule and history boxes, new column widths for stats and map, and moving notes into new row.

### DIFF
--- a/lib/bike_brigade_web/live/rider_live/show.html.heex
+++ b/lib/bike_brigade_web/live/rider_live/show.html.heex
@@ -20,20 +20,20 @@
         </div>
         <div class="">
             <h1 class="mb-4 text-3xl font-medium leading-none sm:text-4xl"><%= @rider.name %> <small class="text-sm"><%= @rider.pronouns %></small></h1>
-            <div class="flex flex-col justify-start text-sm leading-6 sm:text-base md:flex-row md:items-center">
-                <div><Heroicons.Outline.clipboard_list aria-label="Calendar" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 sm:mr-1" />Signed up on: <%= format_date(@rider.signed_up_on) %></div>
-                <div><Heroicons.Outline.fire aria-label="Fire" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 md:ml-4 sm:mr-1" />Latest activity: <.latest_campaign_info socket={@socket} rider={@rider} /></div>
-            </div>
-            <div class="flex flex-col justify-start text-sm leading-6 sm:text-base md:flex-row md:items-center">
-                <div><Heroicons.Outline.location_marker aria-label="Marker" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 sm:mr-1" /><%= @rider.address %></div>
-                <div><Heroicons.Outline.phone aria-label="Phone" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 md:ml-4 sm:mr-1" /><%= @rider.phone %></div>
-                <div><Heroicons.Outline.mail aria-label="Mail" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 md:ml-4 sm:mr-1" /><%= @rider.email %></div>
-            </div>
-            <div class="flex flex-col justify-start text-sm leading-6 sm:text-base md:flex-row md:items-center">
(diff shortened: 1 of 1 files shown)


lib/bike_brigade_web/live/rider_live/show.html.heex
```diff
@@ -20,20 +20,20 @@
         </div>
         <div class="">
             <h1 class="mb-4 text-3xl font-medium leading-none sm:text-4xl"><%= @rider.name %> <small class="text-sm"><%= @rider.pronouns %></small></h1>
-            <div class="flex flex-col justify-start text-sm leading-6 sm:text-base md:flex-row md:items-center">
```
mveytsman, line 23: I switched these from `div`s to `p`s but i guess `p` makes more sense here

lib/bike_brigade_web/live/rider_live/show.html.heex
```diff
@@ -20,20 +20,20 @@
...
+                <span class="block md:inline-block"><Heroicons.Outline.phone aria-label="Phone" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 md:ml-2 sm:mr-1" /><%= @rider.phone %></span>
+                <span class="block xl:inline-block"><Heroicons.Outline.mail aria-label="Mail" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 xl:ml-2 sm:mr-1" /><%= @rider.email %></span>
+            </p>
+            <p class="text-sm leading-6 lg:text-base">
+                <span class="block md:inline-block"><Heroicons.Outline.briefcase aria-label="Briefcase" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 sm:mr-1" />
                     <%= live_redirect @rider.capacity, to: Routes.rider_index_path(@socket, :index, %{capacity: [@rider.capacity]}), class: "link" %>
-                </div>
-                <div class="flex items-center"><Heroicons.Outline.tag aria-label="Tag" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 md:ml-4 sm:mr-1" />
+                </span>
+                <span class="block md:inline-block"><Heroicons.Outline.tag aria-label="Tag" class="inline-flex flex-shrink-0 w-5 h-5 text-gray-400 md:ml-2 sm:mr-1" />
```
mveytsman, line 36: This is not wrapping the tags right
Screen Shot 2022-02-09 at 2 04 26 PM

Screen Shot 2022-02-09 at 2 04 39 PM

mveytsman (commented):
Left a comment inline about tag display.

The schedule is taking on the same height as the history (and both are too tall I think). This is what it looks like when you have 1 thing in the history and nothing scheduled

Screen Shot 2022-02-09 at 2 07 11 PM

I moved the notes below the schedule to keep it on the page for riders that have a lot of history. Otherwise it will be pushed all the way down with the scroll ... here's what it looks like now if you did a lot of programs:

Screen Shot 2022-02-09 at 2 16 48 PM


What do you think?

chadmohr:
1. Height of boxes: I set a fixed height which is for sure not ideal. It is not possible in any way I can conceive of to have the the same height, since they are inner divs. I suppose I could set a min and max height, but I was hoping never to have to scroll the schedule. I will ponder this more.
2. I noticed the tags weren't working too. I will fix this.
3. The notes in the same column as the schedule makes the height of schedule/history a major pain. If it is important maybe it moves up somewhere?

mveytsman:
> Height of boxes: I set a fixed height which is for sure not ideal. It is not possible in any way I can conceive of to have the the same height, since they are inner divs. I sup
pose I could set a min and max height, but I was hoping never to have to scroll the schedule. I will ponder this more.

Oh I see now! In my testing the history and schedule wasn't scrolling so I didn't notice that was the intention - I think the `overflow-x-auto` should be an `overflow-y-auto` and they may be on the wrong elements? Do we want per column scroll on the schedule or just the whole thing?

chadmohr:
@mveytsman I took another kick at the can. Its better, though not perfect

chadmohr:
I figured it out :-). Forgive me if I am making a mess if this Git process

image

chadmohr:
The latest commit hopefully addresses some small spacing issues in the rider details

image
image

mveytsman (approved):
I think the `!leading-relaxed` is right thanks for figuring this all out!

PR #63, opened 2022-02-08
CSS fixes to campaign assignments screen

chadmohr:
Some CSS fixes to make the assignments screen work on mobile. Riders/tasks/map columns and also left and right button group columns stack on mobile screen widths. Note that the Tailwinds ordering classes appear to not be working, so the map is currently always last.

Also note that the button are a bit intense, but that seems a large fix for now.

image


mveytsman (approved):
We got it to work correctly with `order-last` together!

PR #64, opened 2022-02-08
CSS fix to campaign list

chadmohr:
Make campaign name and buttons stack on mobile to prevent content from being wider than screen

image


mveytsman (approved):
👍

PR #81, opened 2022-03-30
Add text based itineraries

jennaleeb:
For riders that do not have a smartphone, or do not have data capabilities, provide a way for them to receive text-only instructions as a follow-up message.

image

image


mveytsman (approved):
Couple of minor suggestions about code and copy, otherwise  :shipit:

Issue #94, opened 2022-04-28
Migrate devcontainer to a straight docker image


mveytsman:
Currently the devcontainer installs nix onto a debian machine. This is a problem because

1) Nix can be slow inside a container
2) Launching a new container takes a long time
3) VSCode has to use the nix environment selector otherwise the language server doesn't work. This is buggy as heck, and will not support flakes.
4) We use the same versions of things inside both the devcontainer and when developing locally, but we still have a disconnect with production which uses a different Docker container

It's possible to get nix/flakes to spit out a docker image for us - but then we would have to figure out how to ship the *image* to codespaces which expects Docker *files* not Docker *images*.

I do think the above is the ideal state of things, but it's also very cutting edge and would take a lot of effort to get there. As a short/medium term solution let's stop messing about with putting userland nix inside docker containers and instead build a devcontainer as a docker image directly.

We can follow the Dockerfile we have for deploying to production. Ideally we would share the same file (or rather the devcontainer docker image is stage 1 of the 2 stage production docker). Baring that, we should at least use the same versions of elixir/erlang. If they have to be separate files, they should at least share a config option with the same name(s).

mveytsman:
Note: sourcing env.local is done inside the [flake.nix](https://github.com/bikebrigade/dispatch/blob/456442668df0efd13477a247c8332d1645db5dfa/flake.nix#L44-L47). Creating the file and other setup stuff are done in the [postcreate]https://github.com/bikebrigade/dispatch/blob/456442668df0efd13477a247c8332d1645db5dfa/.devcontainer/devcontainer.json#L43).

Both need to happen for the dev environment. I'm not sure how env var configuration is intended to work in devcontainers -- this nix hook relies on the shell being started. You could source the env file in the Dockerfile, or perhaps there's a happy path Github intends you to take here.

ShayMatasaro:
ill start working on it

ShayMatasaro:
the project root already has the docker-compose and dockerfile that are used in prod , ill switch the devcontainer to use those , instead of using a separate dev config.

note 
:if we want the dev env to have more features (Ubuntu instead of alpine, ZSH, etc...), separate configs will be needed

mveytsman:
> the project root already has the docker-compose and dockerfile that are used in prod , ill switch the devcontainer to use those , instead of using a separate dev config.

I found that I had to do this for the devcontainer to launch postgres right: 
[...]

PR #107, opened 2022-04-30
removed nix and switched to a simple docker

ShayMatasaro:
https://github.com/bikebrigade/dispatch/issues/94

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -40,5 +40,10 @@
 	// Uncomment to connect as a non-root user if you've added one. See https://aka.ms/vscode-remote/containers/non-root.
 	"remoteUser": "vscode",
 	"userEnvProbe": "loginInteractiveShell",
+
+	//create .env.local
+	"initializeCommand": "cp .env.local.sample .env.local",
+
+
 	// "postCreateCommand": ["/bin/bash", "-lic", "--", "nix-shell --run 'cp .env.local.sample .env.local && mix local.hex --force && mix local.rebar --force && mix do deps.get, ecto.create, ecto.migrate, run priv/repo/seeds.exs && npm install --prefix=assets/'"]
 }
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -40,5 +40,5 @@
 	// Uncomment to connect as a non-root user if you've added one. See https://aka.ms/vscode-remote/containers/non-root.
 	"remoteUser": "vscode",
 	"userEnvProbe": "loginInteractiveShell",
-	"postCreateCommand": ["/bin/bash", "-lic", "--", "nix-shell --run 'cp .env.local.sample .env.local && mix local.hex --force && mix local.rebar --force && mix do deps.get, ecto.create, ecto.migrate, run priv/repo/seeds.exs && npm install --prefix=assets/'"]
+	// "postCreateCommand": ["/bin/bash", "-lic", "--", "nix-shell --run 'cp .env.local.sample .env.local && mix local.hex --force && mix local.rebar --force && mix do deps.get, ecto.create, ecto.migrate, run priv/repo/seeds.exs && npm install --prefix=assets/'"]
```
mveytsman, line 43: We still to run the post creation stuff on startup. See  #110 for making it simpler.

mveytsman:
Good start!

As discussed we will want to load the env file before we run anything (including the db migrations). Or put some default env vars into the docker container.

See how it's done in nix here: https://github.com/bikebrigade/dispatch/blob/main/flake.nix#L44-L46

ShayMatasaro:
.env.local added to docker

.devcontainer/docker-compose.yml
```diff
@@ -8,6 +8,8 @@ services:
     build:
       context: .
       dockerfile: .devcontainer/Dockerfile
+    env_file:
```
mveytsman, line 11: Oh cool, I didn't know about `env_file:`

I think that we will want `initializeCommand: "cp .env.local.sample .env.local"` in the devcontainer.json -- .env.local wouldn't exist yet and be gitignored...

ShayMatasaro: i tested "initializeCommand: "cp .env.local.sample .env.local""  and it looks like it runs locally , not in the docker container , which is an issue on Windows.

ill move it to the dockerfile  and test

ShayMatasaro: adding the cp in dockerfile doesnt work , because at that stage the workspace dir is not  mapped 

putting it in devcontainer.json(onCreateCommand, etc....) also does not work , because docker-compose runs before and fails if .env.local is not present

PR #125, opened 2022-06-03
Add Programs and Campaigns to seeds

### LINKED ISSUE #122: Add Programs and Campaigns to seeds
- We should have at least two programs with at least one campaign each
- Programs should have items
- Campaigns should have a few tasks, and maybe some riders assigned to them
- These should be idempotent like the current seeds
- It would be nice if the campaign's date is the day you run the seeds file.

mariusbutuc:
Screen Shot 2022-06-08 at 11 39 49 AM

Resolves #122.

### DIFF
--- a/lib/bike_brigade/delivery.ex
+++ b/lib/bike_brigade/delivery.ex
@@ -677,9 +677,20 @@ defmodule BikeBrigade.Delivery do
       iex> list_items()
       [%Item{}, ...]
 
+      iex> list_items(42)
+      [%Item{program_id: 42, …}, …]
+
   """
-  def list_items do
+  def list_items(program_id \\ nil) do
     query = from i in Item, order_by: i.name
+
+    query =
+      if program_id do
+        query
+        |> where([i], i.program_id == ^program_id)
+      else
+        query
+      end
 
     Repo.all(query)
     |> Repo.preload(:program)
--- a/priv/repo/seeds/seeder.ex
+++ b/priv/repo/seeds/seeder.ex
@@ -7,11 +7,10 @@ defmodule BikeBrigade.Repo.Seeds.Seeder do
     Utils
   }
 
-  alias BikeBrigade.Accounts.User
   alias BikeBrigade.Repo.Seeds.Toronto
 
   def user() do
-    {:ok, %User{} = user} =
+    {:ok, %Accounts.User{} = user} =
       Accounts.create_user(%{
         name: "Dispatcher McGee",
         phone: "647-555-5555",
@@ -22,11 +21,13 @@ defmodule BikeBrigade.Repo.Seeds.Seeder do
   end
 
   def program() do
-    {:ok, %Delivery.Program{}} =
+    {:ok, %Delivery.Program{} = program} =
       Delivery.create_program(%{
         name: Faker.Company.name(),
         start_date: LocalizedDateTime.now()
       })
+
+    program
   end
 
   def campaign_for_program(program) do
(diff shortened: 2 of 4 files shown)


lib/bike_brigade/delivery.ex
```diff
@@ -677,10 +677,21 @@ defmodule BikeBrigade.Delivery do
...
+  def list_items(program_id \\ nil) do
     query = from i in Item, order_by: i.name
 
+    query =
+      if program_id do
+        query
+        |> where([i], i.program_id == ^program_id)
+      else
+        query
+      end
```
mariusbutuc, line 693: Do we prefer to filter down by program id here, or rather leave it up to the caller, similar to [how campaigns are filtered below](https://github.com/bikebrigade/dispatch/pull/125/files#diff-c47a1aff9bf95e01794823e9077bbcf6c1b6321beadf2ace8ee9496415766c88R58)?

priv/repo/seeds/seeder.ex
```diff
@@ -0,0 +1,66 @@
...
+  def rider_for_campaign(campaign) do
+    rider =
+      Riders.list_riders()
+      |> Enum.random()
+
+    # validate_required([:campaign_id, :rider_id, :rider_capacity, :enter_building, :token])
+    Delivery.create_campaign_rider(%{
+      campaign_id: campaign.id,
+      rider_id: rider.id
+    })
```
mariusbutuc, line 44: Should the other fields be required for seeded data as well?

priv/repo/seeds/seeder.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  def campaign() do
+    current_week =
+      LocalizedDateTime.today()
+      |> Date.beginning_of_week()
+
+    {:ok, campaign} =
+      Delivery.create_campaign(%{
+        # key :name not found in: %BikeBrigade.Delivery.Campaign{…}
+        name: Faker.Company.name(),
```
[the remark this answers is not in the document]
mariusbutuc: Oops, the issue stems from here: https://github.com/bikebrigade/dispatch/blob/eb622fe4f6dab7c917d678d3d7a322a01f97da44/lib/bike_brigade_web/views/campaign_helpers.ex#L98-L106 especially line 104, which tells me the `campaign.program` relation is not set up correctly… debugging further.

priv/repo/seeds/seeder.ex
mveytsman: I think if the fields are not there this would fail right?

mveytsman: nevermind they have defaults :)

priv/repo/seeds/seeder.ex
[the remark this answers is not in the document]
mariusbutuc: Resolved in #126.

priv/repo/seeds/seeder.ex
```diff
@@ -0,0 +1,113 @@
...
+
+    rider
+  end
+
+  def task_for_campaign(campaign) do
+    {:ok, %Delivery.Task{} = task} =
+      campaign
+      |> Delivery.create_task_for_campaign(%{
+        dropoff_name: Faker.Person.first_name(),
+        dropoff_location: Toronto.random_location()
```
mariusbutuc, line 88: In order to create a task (for a campaign) it seems that the drop-off name and location are required. Does this match "real life"? 

If so, thoughts on adding a `validate_required/3` for them to shorten the feedback loop and communicate this detail?

mveytsman (approved):
:shipit:

PR #146, opened 2022-07-11
Add timezone to the format string

### LINKED ISSUE #86: Show timezone in scheduled messages
A dispatcher is currently on vacation abroad and it's not clear wether scheduling is in EST or her local timezone. 

Screen Shot 2022-04-05 at 11 04 04 AM

We've had issues with using the computer's timezone before. We should add the timezone to how the date is printed when showing you when it was scheduled and when it was last sent.

Also this should be updated in the index view
Screen Shot 2022-04-05 at 11 09 58 AM

nathan-louie:
Closes #86 
image


mveytsman (approved):
:shipit: 👍

PR #152, opened 2022-07-15
Fix bug with local storage adapter and make adapter config happen at runtime

mveytsman:
This PR does two things

1) If you don't have any google config we use the local storage adapter instead of the Fake adapter for media storage - this lets us upload photos to send when testing in dev mode
2) The adapter configs use module variables which get set at compile time. This caused my env to enter a weird state because when I changed the dev.ex, media_storage.ex wasn't recompiled... I followed the guidelines in https://hexdocs.pm/elixir/1.13/library-guidelines.html#avoid-compile-time-application-configuration to switch to using a function so we can set the adapter at runtime

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -146,8 +146,8 @@ case System.get_env("GOOGLE_SERVICE_JSON") do
 
   _no_json ->
     config :bike_brigade, :media_storage,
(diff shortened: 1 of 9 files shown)


mveytsman:
@nathan-louie this will fix the bug you were having.

PR #163, opened 2022-07-19
Add week query params to URL in campaign

### LINKED ISSUE #151: Expose query string in URL for campaigns/index page
Right now, as you navigate across weeks in the Campaigns/Index page, the URL doesn't change. We update an internal state variable (`current_week`), instead we should keep it in the query params, i.e. `/campaigns/?current_week=2022-01-01`).

nathan-louie:
image

Now can see the current week as a param in the URL of the campaign page.

Closes #151


mveytsman (approved):
LGTM

PR #169, opened 2022-07-21
Get chromedriver via nix

mveytsman:
We used to download chromedriver via a script because of some nixos issues. I believe that pulling chromedriver directly in our flake.nix should work fine - instead of trying to tie it to the version of Chrome we already have installed.

- [ ] Test in CI
- [x] Test on Darwin
- [x] Test on NixOS


mveytsman:
NixOS confirmed by @shazow

Issue #136, opened 2022-07-06
Rider itinerary page (basic)


mveytsman:
A Rider should be able to see all of the deliveries they are signed up for.

Right now we have "Delivery details pages" that work for a single delivery. 

We would like an "itinerary" that gives you a list of these ordered by day.

The "basic" version can just aggregate data (e.g. "You are delivering 10 items for 5 campaigns"), and link to the existing iteneraries.

Also note that we won't show the rides later this week to people because of how we dispatch - so we want some "Beta" copy.

chadmohr:
This is super basic, but its a start - https://xd.adobe.com/view/cbf01281-7c58-4aa0-b403-a9ef063e064c-9ea3/

mveytsman:
@nathan-louie to take this on this week 

- for testing, you can create a couple campaigns on the same day and assign a given rider to deliveries
- We don't have a rider per user yet , so we can call it "My profile" for testing but keep a hardcoded rider id
- Right now the delivery pages are all public with a random token in the URL, we will want to put them behind login, with a more reasonable url.

mveytsman:
As discussed:

- add some buttons to browse daily iteneraries by date.
- @mveytsman to ship enough of #135 to enable @nathan-louie to write integration tests.

chadmohr:
I am now recalling that previously we discussed how it may confuse riders to see itineraries for additional days because we generally only dispatch one day in advance. So riders would very likely end up seeing empty future days when in fact they have signe dup for something.

mveytsman:
I remember this conversation.

I feel like since we're shipping this with a slow rollout we can mitigate it with some copy / block you from going into the future if need be.

Still think that's it's worth to have as we test/roll things out...

mveytsman:
@teesloane to 🚀 this

- [ ] Integration tests around it
- [ ] Improve copy around how it's beta 
- [ ] Add to sidebar for Riders
- [ ] Make it the default page when you log in as a rider

PR #174, opened 2022-07-30
Create itinerary page with basic features

nathan-louie:
image

Works on #136


lib/bike_brigade_web/live/itinerary_live/index.ex
```diff
@@ -0,0 +1,54 @@
...
+  defp get_location(assigns) do
+    ~H"""
+    <div class="mt-2 sm:flex sm:justify-between">
+      <div class="sm:flex">
+        <p class="flex items-center text-sm text-gray-500">
+          <Heroicons.Outline.location_marker
+            aria-label="Location"
+            class="flex-shrink-0 mr-1.5 h-5 w-5 text-gray-400"
+          />
+          <%= Delivery.get_campaign(assigns.delivery.id).location.address %>
```
mveytsman, line 48: Note that `assigns.delivery.id` can be `@delivery.id` in here.

lib/bike_brigade_web/live/itinerary_live/index.ex
```diff
@@ -0,0 +1,54 @@
...
+        if not is_nil(socket.assigns.current_user.rider_id),
+          do: socket.assigns.current_user.rider_id,
+          else: 26
+      ),
+      # today
+      ~D[2022-06-27]
+    )
+  end
+
+  defp get_location(assigns) do
```
mveytsman, line 39: I like that you named this function `get_location` since it is in fact getting something from the database, it's good to have functions give you that kind of clue from the name.

This also is a clue that we want to split up the concerns - when you have a " function component" -  a function that's returning HTML, we want to keep it [pure](https://en.wikipedia.org/wiki/Pure_function) - that is just rendering the input.

I think what you want to do is something like `@delivery.location.address` or `@campaign.location` if you rename as suggested above.

If you try that now you'll see that the location is missing - it will an association that isn't loaded. We can use `Repo.preload` on the campaign inside of `fetch_deliveries`, but better yet would be to update `Riders.list_campaigns_with_task_counts` to include that preload

mveytsman (changes requested):
@nathan-louie this is looking great!

I left some feedback with general and specific suggestions. We can spend some time going over it tomorrow!

lib/bike_brigade_web/live/components/sidebar.ex
```diff
@@ -26,6 +26,15 @@ defmodule BikeBrigadeWeb.Components.Sidebar do
...
       </.sidebar_link>
+      <.sidebar_link
+        selected={@current_page == :itinerary}
+        to={Routes.itinerary_index_path(@socket, :index)}
+      >
+        <:icon let={class}>
+          <Heroicons.Outline.clipboard_list class={class} />
+        </:icon>
+        My Itinerary
+      </.sidebar_link>
```
nathan-louie, line 37:
```suggestion
```

mveytsman:
Addresses #174

lib/bike_brigade_web/live/itinerary_live/index.html.heex
```diff
@@ -0,0 +1,90 @@
...
+              </div>
+              <.get_location campaign={campaign} />
+              <div class="mt-2 sm:flex sm:justify-between">
+                <div class="sm:flex">
+                  <p class="flex items-center text-sm text-gray-500">
+                    <Heroicons.Outline.shopping_bag
+                      aria-label="Location"
+                      class="flex-shrink-0 mr-1.5 h-5 w-5 text-gray-400"
+                    />
+                    <%= task_count %> Food <%= Inflex.inflect("Hamper", task_count)%>
```
mveytsman, line 79: We can't always assume it's a food hamper

lib/bike_brigade/riders.ex
```diff
@@ -139,18 +139,15 @@ defmodule BikeBrigade.Riders do
     query =
       from c in Campaign,
         as: :campaign,
-        join: cr in CampaignRider,
```
mveytsman: Nice! Thanks for catching this and bringing back the old code.

I think that we could probably refactor this entire function out and use `get_itinerary` in the place where this is used BUT this is out of the scope of this PR.

lib/bike_brigade_web/live/itinerary_live/index.ex
```diff
@@ -0,0 +1,72 @@
...
+
+    if rider_id do
+      socket
+      |> assign(
+        :campaign_riders,
+        Riders.get_itinerary(rider_id, date)
+      )
+    else
+      socket
+      |> assign(:campaigns, [])
```
mveytsman, line 45:
```suggestion
      |> assign(:campaign_riders, [])
```

I missed updating this, which actually makes one of our tests fail, so good to have test :)

mveytsman (approved):
This looks like it's working. I'd like to see a "happy path test" for showing campaigns, but that can come later / we can work on it together.

Issue #177, opened 2022-08-04
Option to hide items when editing a program


nathan-louie:
![image](https://user-images.githubusercontent.com/53024905/182922557-11a29345-a810-4dd9-b813-3d8d81f30898.png)

mveytsman:
Noting that we may actually want to hide the item instead of delete it, as to not mess up previous stat around items.

mveytsman:
Steps:

- [ ] Add a hidden field to `Item`
- [ ] Show/edit hidden state when editing items
- [ ] Hide hidden items when showing items in the delivery page

mveytsman:
Closed in #179

PR #179, opened 2022-08-05
Able to hide/soft-delete items in a program

nathan-louie:
Works on #177

image

I noticed that when it is not hidden, there's a gray border that is shown, and it seems like there's a white box that is visible/misaligned? (Should fix)

### DIFF
--- a/lib/bike_brigade_web/live/program_live/form_component.html.heex
+++ b/lib/bike_brigade_web/live/program_live/form_component.html.heex
@@ -152,9 +152,11 @@
             <:td let={item}>
               <%= item.category %>
             </:td>
-            <:td let={item} class="flex justify-center">
+            <:td let={item}>
               <%= if item.hidden do %>
-                <Heroicons.Outline.eye_off class="flex-shrink-0 w-6 h-6 mx-1 text-gray-600" />
+                <div class="flex justify-center">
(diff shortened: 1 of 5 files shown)


lib/bike_brigade_web/live/program_live/form_component.html.heex
```diff
@@ -147,6 +152,11 @@
             <:td let={item}>
               <%= item.category %>
             </:td>
+            <:td let={item} class="flex justify-center">
+              <%= if item.hidden do %>
+                <Heroicons.Outline.eye_off class="flex-shrink-0 w-6 h-6 mx-1 text-gray-600" />
+              <% end %>
+            </:td>
```
mveytsman, line 159:
```suggestion
            <:td let={item}>
              <%= if item.hidden do %>
                <div class="flex justify-center">
                  <Heroicons.Outline.eye_off class="flex-shrink-0 w-6 h-6 mx-1 text-gray-600" />
                </div>
              <% end %>
            </:td>
```

mveytsman: I steered you wrong with adding `flex` to the td cell, instead we should have made a flex / justify-center container around the icon.

mveytsman (approved):
Looks great!

I added a suggestion to fix that border, otherwise 🚢 it!

PR #187, opened 2022-08-17
update copy for rider search

jennaleeb:
I always forget you can search by number, helpful as the number of duplicate first names is getting larger!


mveytsman (approved):
🎉

PR #204, opened 2022-12-08
Add: DotenvParser

teesloane:
Moves environment variable loading to runtime when in :dev rather than rely on the nix shell to do this for us.


mveytsman (approved):
Thanks!!

PR #205, opened 2022-12-08
Add: mobile friendly flash message

### LINKED ISSUE #203: Improve disconnected errors (on mobile)
This ticket is about two things

1) We're going to be disconnected when the app is backgrounded on mobile. Right now, our error is a little too aggressive / jarring. We should make it say something like "Reconnecting" and be a warn instead of a red error. 

image

See [this comment from Chris](https://news.ycombinator.com/item?id=28623250) for more context

2) The flash error above is too wide on mobile, we should make it narrower. It's from the Phoenix default templates, and they may have changed the styling since I last looked...

teesloane:
- add "warn" @kind for flash
- puts title and flash inner_block onto a single row when on mobile.

closes #203 

https://user-images.githubusercontent.com/12987958/206572956-aa738c08-63ad-4d8a-91dc-aee42734837c.mp4

### DIFF
--- a/lib/bike_brigade_web/components/layouts/app.html.heex
+++ b/lib/bike_brigade_web/components/layouts/app.html.heex
@@ -155,9 +155,11 @@
           <.flash
             id="disconnected"
             kind={:warn}
-            title="You're Offline."
-            close={true}
-            autoshow={true}
+            title="Offline."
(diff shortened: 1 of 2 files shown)


mveytsman (approved):
🎉

PR #212, opened 2022-12-15
Fix: dotenv works when running mix test locally

teesloane:
As in #204, we also need to load environment variables for running tests.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -7,7 +7,10 @@ case config_env() do
       config :bike_brigade, BikeBrigade.Repo, hostname: "postgres"
     end
 
-  DotenvParser.load_file(".env.local")
+    DotenvParser.load_file(".env.local")
+
+  :test ->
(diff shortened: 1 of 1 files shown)


teesloane:
Of course this doesn't work in ci 😹 . I'll have to think of something else to get this working.

mveytsman:
@teesloane do tests fail locally for you without loading the env vars?

... they run fine for me, and I don't think we have any env vars we need to load on CI....

teesloane:
Yeah, the env vars weren't loading for me. 

<detail>
<summary> Here's the output</summary>
</detail>

mveytsman:
@teesloane what do you think of https://github.com/bikebrigade/dispatch/pull/215 as the solution?

teesloane:
> @teesloane what do you think of #215 as the solution?

checked it out and it compiles and runs the tests properly 👍 . Let's go for it.

mveytsman:
Closing in favor of #215

PR #213, opened 2022-12-16
Explore adding tests for riders index

teesloane:
This is a draft PR on exploring the start of what additional live view tests could look like. I'm not sure _how_ valuable testing certain functionality is for LiveView as I don't want to make refactoring too much of a pain where it might feel like you have to maintain a lot of tests as you go (or end up deleting them.)

I used the riders page as an example demo since it seems straight-foward. The test I added:

- updated the `table` core component to be able to take a row_class, so that we can...
- enable getting those values  with Floki (so, technically could also use a data-test-* html attribute instead of a class if desired)
- wrote a test that checks that pagination is working as expected.

Some other example tests I imagine would be useful:

1. test that multiple users can be selected and then when navigating to "bulk message" they are all listed in the modal
1. test that clicking on a user takes you to the user page
1. test the table can sort properly by sortable headings 

Let me know if you think this would be useful!

### DIFF
--- a/test/bike_brigade/location_test.exs
+++ b/test/bike_brigade/location_test.exs
@@ -87,7 +87,7 @@ defmodule BikeBrigade.Locations.LocationTest do
                |> Ecto.Changeset.apply_action(:update)
     end
 
-    test "change postal", %{location: %{postal: postal, coords: coords} = location} do
+    test "change postal", %{location: %{postal: postal, coords: coords} = _location} do
       assert {:error, _} =
                Location.change_location(%Location{}, :postal, "m62aaaa")
                |> Ecto.Changeset.apply_action(:update)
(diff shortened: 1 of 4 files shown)


test/bike_brigade_web/live/rider_live_test.exs
```diff
@@ -6,11 +6,23 @@ defmodule BikeBrigadeWeb.RiderLiveTest do
   describe "Index" do
     setup [:create_rider, :login]
 
-    test "lists all riders", %{conn: conn, rider: rider} do
-      {:ok, _index_live, html} = live(conn, ~p"/riders")
+    test "lists all riders, with working next/previous pagination", %{conn: conn} do
+      # REVIEW: is there a way I can just call :create_rider here? Do I have to import it from conn_case?
```
mveytsman, line 10: It's imported already as part of `use BikeBrigadeWeb.ConnCase` 

It just expects a context map as an argument (see https://hexdocs.pm/ex_unit/1.12/ExUnit.Callbacks.html).

Note that ` use BikeBrigadeWeb.ConnCase` also imports the fixture, so you could just do `for _ <- 0..25, do: fixture(:rider)`

mveytsman (commented):
Looks good so far! I think this is a great direction, and the row-class is useful!!

teesloane:
Ready for re-review

test/bike_brigade/location_test.exs
```diff
@@ -96,13 +96,13 @@ defmodule BikeBrigade.Locations.LocationTest do
                Location.change_location(%Location{}, :postal, "m6a")
                |> Ecto.Changeset.apply_action(:update)
 
-      assert {:ok, %{postal: postal, coords: ^coords}} =
+      assert {:ok, %{postal: ^postal, coords: ^coords}} =
```
mveytsman, line 99: good catch!

mveytsman (approved):
Sorry I missed this.

Looks great 🚀 !

PR #218, opened 2022-12-22
Search in the conversations pane

mveytsman:
Added a search bar to the conversations pane.,

Right now we just hide elements in the front end using a JS hook.

Screen Shot 2022-12-22 at 12 34 47 PM

Should be rebased against #217 when that lands to move our hook into it's own file

### DIFF
--- a/assets/js/hooks/conversation_list.js
+++ b/assets/js/hooks/conversation_list.js
@@ -24,6 +24,25 @@ const ConversationList = {
         this.el.prepend(msg)
       }
     });
+
+    this.handleEvent("only_show", ({
+      ids
+    }) => {
+      Array.from(this.el.children).forEach((item) => {
+        if (ids.includes(parseInt(item.dataset.riderId))) {
+          item.classList.remove("hidden");
+        } else {
+          item.classList.add("hidden");
+        }
+      });
+    });
+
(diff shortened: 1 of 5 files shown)


assets/js/app.js
```diff
@@ -268,6 +268,25 @@ Hooks.ConversationList = {
...
+      Array.from(this.el.children).forEach((item) => {
+        if (ids.includes(parseInt(item.dataset.riderId))) {
+          item.classList.remove("hidden");
+        } else {
+          item.classList.add("hidden");
+        }
+      });
+    });
+
+    this.handleEvent("clear_search", ({}) => {
```
teesloane, line 284:
```js
    this.handleEvent("conversation_list:clear_search", ({}) => {
```
Thoughts on namespacing events? I'm not sure that we'll run into clashing events, but couldn't hurt (unless the events are somehow scoped to the hook object, I don't know?)

teesloane (commented):
Looks good so far. My tests are all passing locally, so I don't know what the problem is on CI.

assets/js/app.js
mveytsman: Oh yeah that's a great idea!

teesloane (approved):

PR #223, opened 2023-01-17
Edit items inside forms

### LINKED ISSUE #101: Programs/new: Item selection is not shown when you create the program, only on edit
New
Screen Shot 2022-04-29 at 11 06 16 AM
Edit
Screen Shot 2022-04-29 at 11 09 01 AM

mveytsman:
Closes #101 

Adds inline editing for items to program form:
Screen Shot 2023-01-17 at 3 43 17 PM


lib/bike_brigade_web/live/program_live/form_component.html.heex
```diff
@@ -74,13 +73,50 @@
...
+              phx-click={JS.push("remove_item", value: %{index: i}, target: @myself)}
+              type="button"
+            >
+              <span class="sr-only">Close</span>
+              <Heroicons.x_mark solid class="w-6 h-6" />
+            </.button>
+          </:action>
+        </.table>
+
+        <%!--<.table id="items" rows={@program.items}>
```
teesloane, line 119: Do we want to leave this commented out or just delete it?

lib/bike_brigade_web/live/program_live/form_component.html.heex
```diff
@@ -74,13 +73,50 @@
...
+            <.input
+              type="select"
+              field={{item_form, :category}}
+              options={Ecto.Enum.values(Delivery.Item, :category)}
+            />
+          </:col>
+
+          <:col :let={{item_form, i}} label="Hidden?">
+            <.input type="checkbox" field={{item_form, :hidden}} />
+          </:col>
```
teesloane, line 103: I'm not super clear on why this can be hidden — do you think it might confuse users as well?

teesloane (approved):
Looking good! Added a comment or two, but also approved :)

lib/bike_brigade_web/live/program_live/form_component.html.heex
mveytsman: I changed the label to "archived?" which I think is clearer

PR #227, opened 2023-01-19
Validate location isn't blank

mveytsman:
No more trips to [Null Island](https://en.wikipedia.org/wiki/Null_Island)

### DIFF
--- a/lib/bike_brigade/locations/location.ex
+++ b/lib/bike_brigade/locations/location.ex
@@ -49,10 +49,12 @@ defmodule BikeBrigade.Locations.Location do
     |> cast(params, @fields)
     |> validate_required([:coords, :city, :province, :country])
     |> validate_change(:coords, fn :coords, coords ->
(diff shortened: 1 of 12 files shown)


config/runtime.exs
```diff
@@ -3,11 +3,80 @@ import Config
...
+    end
+
+    case System.get_env("MAILCHIMP_API_KEY") do
+      nil ->
+        config :bike_brigade, :mailchimp, adapter: BikeBrigade.MailchimpApi.FakeMailchimp
+
+      key ->
+        config :bike_brigade, :mailchimp, adapter: BikeBrigade.MailchimpApi.Http
+        config :mailchimp, api_key: key
+    end
```
teesloane, line 79: will a rebase on main remove these from the diff?

mveytsman: Good catch thanks!

teesloane (approved):
it looks .... good to me?

mveytsman:
Closes #227

PR #231, opened 2023-02-03
Fix: enable opt out of slideover phx-click-away

teesloane:
Bug replication:
- visit: http://localhost:4000/campaigns/<id>/messaging
- try and insert an emoji
- the slideover closes

This is a quick fix that allows you to disable the click-away for a slideover via an attribute. Not the best solution, but the user can still press the "X" to exit.


mveytsman:
Can you try a different approach here?

https://emoji-button.js.org/ is the button we use. It looks like it's deprecated in favor of a new version: https://picmojs.com/

I notice that PicMo has a `rootElement` option when creating the button. I think that's the trick! click-away wont register as clicking away if the emoji picker element is a child of the slideover - Can you try switching to PicMo and setting the root element to be the button or something next to it and see if it fixes the issue?

PR #238, opened 2023-06-03
Fix: make sure we can remove tags.

teesloane:
In response to this [bug report](https://torontobikebrigade.slack.com/archives/C016VGHETS4/p1685498283175119).

The problem: when we remove all tags, the live component that handles the tag state doesn't get included in the form_params going into the live_view — even if it's just an empty list.

Hacky fix: use a map merge to make sure we at least have a default empty list for maps.

I feel that there is probably a better solution, but I'm a little fuzzy on how to work with live components that have child state (as in, how should I get that empty list that would [be here](https://github.com/bikebrigade/dispatch/blob/191f1accb69b2c3eabee5357ae4842a043fb6a49/lib/bike_brigade_web/live/rider_live/tags_component.ex#L36-L40) into the parent form?

Also, I added a test in the process of using TDD to try and fix this / figure out what was wrong. It's not really related, but `¯\_(ツ)_/¯`.

Here's a before and after:

https://github.com/bikebrigade/dispatch/assets/12987958/971001c4-aa8a-4b17-b851-6fb63e37d713

### DIFF
--- a/lib/bike_brigade_web/live/rider_live/form_component.ex
+++ b/lib/bike_brigade_web/live/rider_live/form_component.ex
@@ -128,6 +128,7 @@ defmodule BikeBrigadeWeb.RiderLive.FormComponent do
   end
 
   defp save_rider(socket, :edit, rider_form_params) do
+    rider_form_params = Map.merge(%{"tags" => []}, rider_form_params)
     with {:ok, form} <-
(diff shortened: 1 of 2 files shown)


mveytsman (approved):
This looks great thanks <3

PR #246, opened 2023-10-20
Bump dependencies

mveytsman:


mveytsman:
I can't deploy on staging because the staging db is broken (fly issue)

Going to merge anyway

PR #248, opened 2023-10-26
[241] stats polish

### LINKED ISSUE #241: Polish riders profile/stats
- [ ] Validate that the existing code works
- [ ] Make sure that you're only shown things that you have permission for (e.g. I shouldn't be able to add tags to my profile)
- [ ] broken link on size: "small"

Screenshot 2023-10-24 at 3 01 49 PM

teesloane:
Closes #241 

- [?] Validate that the existing code works 
  - Seems fine to me? 
- [X] Make sure that you're only shown things that you have permission for (e.g. I shouldn't be able to add tags to my profile)
- [X] broken link on size: "small"
  - This was broken because the link would take the rider to a dispatch-only page.
  - I have removed this row of items as they are dispatcher specific. 

One issue remains, which is that it seems that riders are unable to actually _save_ their changes in the side `edit` panel when it opens - should this be possible? Or is this a bug?

## Screenshots (After)

Profile _ Bike Brigade Dispatch

CleanShot 2023-10-26 at 15 48 50@2x

### DIFF
--- a/lib/bike_brigade_web/live/rider_live/form_component.html.heex
+++ b/lib/bike_brigade_web/live/rider_live/form_component.html.heex
@@ -35,7 +35,7 @@
       />
     </div>
 
-    <div phx-feedback-for={input_name(f, :tags)}>
+    <div :if={@current_user.is_dispatcher} phx-feedback-for={input_name(f, :tags)}>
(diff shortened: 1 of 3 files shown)


mveytsman (approved):
This looks good! I think we do want to be able to edit our profile. I made this a new issue

https://github.com/bikebrigade/dispatch/issues/249

PR #250, opened 2023-10-27
Add: allow rider to edit themselves.

### LINKED ISSUE #249: Non dispatcher Riders can save changes to profile
Currently they can see the form but not save it

teesloane:
Closes #249.

I noticed that in the routes file, the dispatch route for [editing a rider](https://github.com/bikebrigade/dispatch/blob/6d2c710e843325da2f9839c4da4623b3e8e5f6b5/lib/bike_brigade_web/router.ex#L153) has an action of `:edit`, whereas the route for [riders' action is ](https://github.com/bikebrigade/dispatch/blob/6d2c710e843325da2f9839c4da4623b3e8e5f6b5/lib/bike_brigade_web/router.ex#L103)`:edit_profile`
 
I assumed this was intentional, which is why I created a `save_rider_edit_impl`. It it's not, I can remove most of the code and make sure both pages use the same action.

### DIFF
--- a/lib/bike_brigade_web/live/rider_live/form_component.ex
+++ b/lib/bike_brigade_web/live/rider_live/form_component.ex
@@ -127,21 +127,9 @@ defmodule BikeBrigadeWeb.RiderLive.FormComponent do
     {:noreply, socket}
   end
 
-  defp save_rider(socket, :edit, rider_form_params) do
-    rider_form_params = Map.merge(%{"tags" => []}, rider_form_params)
-    with {:ok, form} <-
-           RiderForm.update_form(socket.assigns.form, rider_form_params),
-         params = RiderForm.to_params(form),
-         {:ok, _rider} <-
-           Riders.update_rider_with_tags(socket.assigns.rider, params, params[:tags]) do
-      {:noreply,
-       socket
-       |> put_flash(:info, "Rider updated successfully")
-       |> push_navigate(to: socket.assigns.navigate)}
-    else
-      _ -> {:noreply, assign(socket, :changeset, socket.assigns.changeset)}
-    end
-  end
+  defp save_rider(socket, :edit, rider_form_params), do: save_rider_edit_impl(socket, rider_form_params)
+  defp save_rider(socket, :edit_profile, rider_form_params), do: save_rider_edit_impl(socket, rider_form_params)
+
 
   defp save_rider(socket, :new, rider_params) do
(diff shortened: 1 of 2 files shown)


mveytsman:
I think the intention was for `:edit_profile` to only let you save the stuff that riders can change, wheras `edit` let's you change anything.

I *think* the way this is writen, a rider could change the `is_dispatcher` flag on themselves if they crafted the request right, which is bad obviously.

Does that seem right? I think the pattern should be an admin changeset and a non admin profile changeset on Rider, and different `Rider.update...` calls for each path. Does that seem right?

teesloane:
> Does that seem right? I think the pattern should be an admin changeset and a non admin profile changeset on Rider, and different `Rider.update...` calls for each path. Does that seem right?

Yes, I think that makes sense. My default would have been to thinking that hiding the fields would be sufficient, but eliminating the possibility of that (whether it be bad actors or just bad luck) makes the most sense. I will look at the code again and let you know if I have any questions.

teesloane:
Updated this so that the changeset for editing of a ride profile differs based on the type of user making the edit.

mveytsman (approved):
I think this works.

To be totally honest I'm a little nervous that the changeset we differentiate is on the `RiderForm` not the `Rider` struct, but I don't have a better way yet -- if we ever get to the `User` modification we should look at this pattern again

teesloane:
@mveytsman I added some tests in this [commit](https://github.com/bikebrigade/dispatch/pull/250/commits/3753bd0dca35d808e62c8bdee0133932d4acc8b9).

Originally, I was going to test the actual trying to change the form, as a rider, and then log in as a dispatcher to make sure it hadn't changed, but it turned out that I couldn't even render the form with the unreachable fields in the first place. So, I added tests to assert that the correct error is raised and tested the inverse (that the admin-only fields only work for admins).

Lmk what you think.

mveytsman:
Ah interesting. It looks like LiveViewTest is not allowing us to submit the form with those fields filled in. So technically this test isn't testing what we want to test - e.g. the error comes from the test framework before it hits our code. So sorry to say this way of testing what we want to test is superflous.

I think it's fine to keep it, but technically it's kind of a test that will always pass if the form doesn't have the field, even if one could change it.

Does that make sense?

teesloane:
hmm, I have some questions about this, but they might be more core to how phoenix works with forms and the problem we are trying to avoid. Let's maybe leave this for now and pair on it when you have a moment?

PR #255, opened 2023-10-31
Clean: remove bogus route.

### LINKED ISSUE #173: Going directly to /riders/ID/edit causes an internal service error
E.g. https://dispatch.bikebrigade.ca/riders/123/edit will break the site.

I think we're forgetting to load something in `mount` that's just being done in `apply_action`?

teesloane:
Fixes #173


mveytsman (approved):
Thank you!

PR #259, opened 2023-11-15
258: Enable editing campaign riders

### LINKED ISSUE #258: Edit rider on campaign page
When a rider is added to a campaign there are a few options - time slot, enter building and capacity
image

It is not possible to edit these without removing the rider and adding them back in. And edit rider option is one solution.

Use case: time is important in bike pirates campaign and the same riders often do it week after week. When Ben duplicates a campaign it carries the rider time slot over, but often the riders choose different times week after week.

teesloane:
Closes #258 

This set of commits makes the following changes:

- add a route to allow the editing of a rider
- update the AddRiderFormComponent to be able to _edit_ riders. 
- rename the AddRiderFormComponent -> CampaignRiderFormComponent
- Update the RiderSelectionComponent to have an update callback (not sure if this is right, but I'm not sure if the same logic is possible in `mount`.)
- Add Tests 

## Screencap of the changes

https://github.com/bikebrigade/dispatch/assets/12987958/b5b7825d-e457-4773-97ac-3ab95997e5fd


lib/bike_brigade/delivery.ex
```diff
@@ -170,6 +170,17 @@ defmodule BikeBrigade.Delivery do
...
 
+  def get_campaign_rider_by_id(campaign_id, rider_id) do
+    CampaignRider |> Repo.get_by(campaign_id: campaign_id, rider_id: rider_id)
+  end
+
+  def update_campaign_rider(%CampaignRider{} = rider, attrs) do
+    rider
+    |> CampaignRider.changeset(attrs)
+    |> Repo.update()
+    |> broadcast(:campaign_updated)
```
mveytsman, line 181: Move closer to create_campaign rider
Note we're broadcasting `:campaign_updated` should be `:campaign_rider_udated`

Update the handle_info to do the same as `:campaign_rider_created`

mveytsman (changes requested):
Love it!

Small change around the broadcast / handling of the update campaign rider event.

teesloane:
Fixed and ready for re-review!

mveytsman (approved):

PR #271, opened 2024-01-25
Add message list to campaigns

### LINKED ISSUE #269: Group message by current campaign
A feature that has been requested, that may be interesting to implement, would be the ability to group the messages pane by riders for a a specific campaign.

In the screenshot below, we could add a "filter by campaign" to the right of the search bar (1) perhaps—clicking it would effectively do a UI group-by in the message page (2)

CleanShot 2024-01-16 at 13 33 19@2x

-----

Alternatively, it could be that on the campaign page for dispatchers, there could be a button (1) that pulls up a drawer from the bottom (2), or a slideover on the right side that has a chat view similar to the `/messages` page - but just the riders for that campaign.

CleanShot 2024-01-17 at 14 47 02@2x

teesloane:
## TL;DR

This PR adds a slideover messaging pane for chatting with riders from a specific campaign. Commits are (relatively) atomic.

Closes #269 

## Screencapture

https://github.com/bikebrigade/dispatch/assets/12987958/eaf96322-b71f-462e-b3c1-35278b7aca97

## Details

- I pitched this feature to dispatchers last week and got some positive feedback and interest
- I created a prototype and iterated on it. Along the way I refactored some now shared code into components and cleaned up the styles on the messages.
- Currently dispatchers are required to a) go to `/messages`, find their rider, and correspond, or b) go to the campaign > click on a rider > click `message`, which will jump them to the chat.
- I have also updated the design of the messages to use a bit more space, as well as add borders between sms-messages.

## To do still:

- [x] write tests
- [ ] get dispatcher input at the next BB meeting on the UI/UX.


teesloane:
@mveytsman Feel free to review now.

lib/bike_brigade/messaging.ex
```diff
@@ -53,6 +53,31 @@ defmodule BikeBrigade.Messaging do
...
   end
 
+
+  @doc"""
+  Used to fetch sms conversations from riders who are riding for a specific
+  campaign. It uses a left_lateral_join (as opposed to the inner_lateral_join in
+  `list_sms_conversastions`) because we need to return all riders from the
+  passed in list, even if they have no sms history (albeit unlikely)).
+  """
+  def list_sms_for_campaign(rider_ids) do
```
mveytsman, line 63: I find this function name confusing.

If it's searching by campaign we should pass it a campaign id and do some more complicated join across the camapaigns table. 

If we want to give it rider ids, we should call it `list_sms_conversations_for_riders` or `list_lasted_sms_messages_for_riders`

Personally i think the campaign ID makes more sense but I haven't thought about it too much

lib/bike_brigade_web/components/sms_message_list_component.ex
```diff
@@ -0,0 +1,139 @@
...
+    ~p"/messages/#{rider_id}"
+  end
+
+  def handle_build_message_link({"/campaign", campaign_id, rider_id}) do
+    ~p"/campaigns/#{campaign_id}/messaging/riders/#{rider_id}"
+  end
+
+
+  # TODO: can :any be a tuple?
+  # TODO: fix indent.
```
mveytsman, line 32: Is there still indenting issue?

lib/bike_brigade_web/components/sms_message_list_component.ex
```diff
@@ -0,0 +1,139 @@
...
+  def handle_build_message_link({"/messages", rider_id}) do
+    ~p"/messages/#{rider_id}"
+  end
+
+  def handle_build_message_link({"/campaign", campaign_id, rider_id}) do
+    ~p"/campaigns/#{campaign_id}/messaging/riders/#{rider_id}"
+  end
+
+
+  # TODO: can :any be a tuple?
```
mveytsman, line 31: I believe so!

mveytsman: We can make a struct here for `RiderMessage` but i think that's overkill for now....

lib/bike_brigade_web/components/sms_message_list_component.ex
```diff
@@ -0,0 +1,139 @@
...
+
+
+  def handle_build_message_link({"/messages", rider_id}) do
+    ~p"/messages/#{rider_id}"
+  end
+
+  def handle_build_message_link({"/campaign", campaign_id, rider_id}) do
+    ~p"/campaigns/#{campaign_id}/messaging/riders/#{rider_id}"
+  end
+
```
mveytsman, line 29: I think these are never used

lib/bike_brigade_web/components/sms_message_list_component.ex
```diff
@@ -0,0 +1,139 @@
...
+  @doc """
+  A helper component to determine where clicking on a rider in an sms list takes us
+  we render the sms list in different capacities; this allows deciding to which sms
+  instance clicking a rider would take us to.
+  """
+  def message_link(assigns) do
+    route = case assigns.sms_list do
+      {"/messages"} -> ~p"/messages/#{assigns.rider_id}"
+      {"/campaigns", campaign_id} -> ~p"/campaigns/#{campaign_id}/messaging/riders/#{assigns.rider_id}"
+    end
```
mveytsman, line 44: This can be better.

We're passing `sms_list` to the component, which seems to only exist to help us determine the url we go to when we click the rider. The name `sms_list` doesn't really point to that, and the shape of the data is unclear - a tuple of a string or a string and number.

I think a better solution is to pass a `rider_link_fn` to the component which would take a `rider_id`, so the caller would make a message list component as either
```
 <.sms_message_list
      rider_link_fn={fn (rider_id) -> ~p"/messages/#{rider_id}" end}
      conversations={@conversations}
```

Or
```
<.sms_message_list
      rider_link_fn={fn (rider_id) -> ~p"/campaigns/#{campaign_id}/messaging/riders/#{rider_id}" end}
      conversations={@conversations}
      
```

Note in the second example the caller knows the campaign_id

lib/bike_brigade_web/components/sms_message_list_component.ex
```diff
@@ -0,0 +1,139 @@
...
+        <%= render_slot(@inner_block) %>
+      </.link>
+    """
+  end
+
+  @doc"""
+  Renders a list of sms messages
+
+  ## Examples
+      <.sms_message_list
```
mveytsman, line 62: Example is missing the `sms_list` (or above `rider_link_fn`)

lib/bike_brigade_web/live/campaign_live/campaign_message_list.ex
```diff
@@ -0,0 +1,50 @@
...
+    rider_ids =
+      assigns.riders
+      |> Map.values()
+      |> Enum.map(& &1.id)
+
+    conversations = Messaging.list_sms_for_campaign(rider_ids)
+
+    rider =
+      cond do
+        # TODO: why did I do this?
```
mveytsman, line 35: I understand the code below but it may be clearer as a nested if or with some comments :)

lib/bike_brigade_web/components/conversation_component.ex
```diff
@@ -0,0 +1,135 @@
...
+        [] -> nil
+      end
+
+    {:ok,
+     socket
+     |> assign(assigns)
+     |> assign_new_sms_message()
+     |> assign(conversation: conversation)
+     |> assign(earliest_timestamp: earliest_timestamp)
+     |> assign(phx_update: "append")}
```
mveytsman, line 38: Let's pair on getting this to use streams! (outside this PR)

mveytsman (changes requested):
Have some feedback and proposition to pair <3

lib/bike_brigade/messaging.ex
teesloane: I see that campaign_id is nullable (and most rows have `campaign_id` set to `null` in my local db) in the `sms_mesages` table. So, I'm thinking that maybe we just change the name for now?

lib/bike_brigade_web/components/sms_message_list_component.ex
teesloane: Fixed!

lib/bike_brigade_web/components/sms_message_list_component.ex
teesloane: Duly noted.

lib/bike_brigade_web/components/sms_message_list_component.ex
teesloane: Correct. Removed.

lib/bike_brigade_web/components/conversation_component.ex
teesloane: Yes, 💯 !

lib/bike_brigade/messaging.ex
```diff
@@ -53,6 +53,31 @@ defmodule BikeBrigade.Messaging do
...
+    query =
+      from r in Rider,
+        as: :rider,
+        left_lateral_join:
+          latest_message in subquery(
```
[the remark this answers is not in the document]
teesloane: Good call, done.

lib/bike_brigade_web/components/sms_message_list_component.ex
teesloane: I thought it wasn't possible to pass lambda's to components, but it turns out I was incorrectly invoking them (forgot you need a `.` -> `rider_link_fn.()`.

Now it works 🎈 !

lib/bike_brigade_web/components/sms_message_list_component.ex
teesloane: Fixed

teesloane:
Think I hit everything. Only [one comment probably needs a quick glance](https://github.com/bikebrigade/dispatch/pull/271#discussion_r1473444206).

test/bike_brigade_web/live/campaign_live_test.exs
```diff
@@ -173,21 +173,20 @@ defmodule BikeBrigadeWeb.CampaignLiveTest do
 
     test "'Text Riders' button is not visible without riders.", %{conn: conn, campaign: campaign} do
       {:ok, view, _html} = live(conn, ~p"/campaigns/#{campaign}")
-      open_browser view
```
mveytsman: 👍

lib/bike_brigade_web/components/sms_message_list_component.ex
```diff
@@ -71,7 +43,10 @@ defmodule BikeBrigadeWeb.Components.SMSMessageListComponent do
         id={"conversation-list-item:#{rider.id}"}
         data-rider-id={rider.id}
       >
-        <.message_link sms_list={@sms_list} rider_id={rider.id}>
+        <.link
```
mveytsman, line 46: Nice I like that you don't even need the message link component anymore

lib/bike_brigade/messaging.ex
```diff
@@ -26,20 +26,24 @@ defmodule BikeBrigade.Messaging do
...
       from r in Rider,
         as: :rider,
-        inner_lateral_join:
-          latest_message in subquery(
-            from m in SmsMessage,
-              where: m.rider_id == parent_as(:rider).id,
-              order_by: [desc: m.sent_at],
-              limit: 1
-          ),
+        inner_lateral_join: latest_message in ^get_latest_message_subquery(),
```
mveytsman, line 46: Small nit but I bumped ecto in #246 and it now gives warnings that the join doesn't specify an `on` condition. 
When you rebase you may need to ad an `on: true` to these joins

mveytsman (approved):

lib/bike_brigade/messaging.ex
teesloane: Done!

teesloane:
Rebased and made a small copy change. Whatever updated, we're getting a lot more compiler warnings when running tests, so I'll make a ticket to fix that.

mveytsman (approved):
👍 LGTM

PR #273, opened 2024-02-06
Simplify:  rider signup table

teesloane:
In this PR, we took on removing the need for a pop-up modal rider sign-up form component. This means making it so that when a rider clicks on the "Sign Up" button, they are immediately signed up and the values for "Pickup Window", "Enter Building" and "Capacity" are set to default values. Whether these values will remain as defaults is to be seen.

During my first iteration of creating the rider signup, there was a user-experience-discomfort with the pop-up modal asking for rider details. For now, we can remove this to make the user experience better, and when the time comes to work with campaigns that require these details to be specified, we can figure a more streamlined approach.

## Before

CleanShot 2024-02-05 at 19 02 51

## After (this PR)

CleanShot 2024-02-05 at 18 53 11

Shout out to @MikeRose151 for making his first contribution!


lib/bike_brigade_web/live/campaign_signup_live/show.ex
```diff
@@ -110,6 +110,30 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Show do
...
 
+  def handle_event("signup_rider", %{"rider_id" => rider_id, "task_id" => task_id}, socket) do
+    %{campaign: campaign, tasks: tasks} = socket.assigns
+    task = Enum.find(tasks, fn task -> task.id == task_id end)
+
+    attrs = %{
+      # this is arbitrary and not actually used by dispatchers anymore.
+      "rider_capacity" => "1",
+      # this will need to be configurable at some point, for campaigns that have pickup time edge cases.
+      "pickup_window" => "no time specififed",
```
mveytsman, line 121: We can make this the pickup window that covers the whole campaign, take a look at e.g. https://github.com/bikebrigade/dispatch/blob/24931a0541fbc08f4ff15f0bb7cf267a4ef48509/lib/bike_brigade_web/live/campaign_signup_live/show.html.heex#L14 for a place we generate that string

mveytsman (changes requested):
Looking good - small nit about the pickup window. We should dump it but for now might as well make it consistent and have a time in there.

mveytsman (approved):
🚀

PR #281, opened 2024-02-29
Add message list to campaigns (take two)

mveytsman:
This is #271 un-reverted with a commit fixing the`defdelegate` issue.

@teesloane I think the problem was just that the `CampaignHelpers` module wasn't aliased, hence the error `function CampaignHelpers.name/1 is undefined (module CampaignHelpers is not available)`

I'm a little bit surprised that this code compiled though with a reference to an unavailable module. Also surprised it worked locally for you @teesloane 

Does it still work with this commit?


teesloane:
> This is #271 un-reverted with a commit fixing the`defdelegate` issue.
> 
> @teesloane I think the problem was just that the `CampaignHelpers` module wasn't aliased, hence the error `function CampaignHelpers.name/1 is undefined (module CampaignHelpers is not available)`
> 
> I'm a little bit surprised that this code compiled though with a reference to an unavailable module. Also surprised it worked locally for you @teesloane
> 
> Does it still work with this commit?

@mveytsman  I checked out the branch and then tested it. `/messages` work as well as the new messages slide over in the campaign view. It did work locally before it crashed on prod. 

If we merge this we should aim to do so during a non-busy time 😓 .

PR #284, opened 2024-03-07
Make campaign signup more mobile friendly

teesloane:
This PR is a rider-portal focused set of changes that makes the campaign sign-up delivery list and campaign sign-up task list more mobile-friendly.

For the Rider Task List, I left the desktop design to continue using a table-based layout and chose to render an unordered list for mobile. This means that the markup shows up twice in the DOM and is hidden by Tailwind. Optionally, I could remove the table entirely on desktop to reduce repetition.

On the implementation side, this does some small refactoring to tidy up how the sign up for campaign button renders as there are lots of permutations based on the rider's state as well as the campaign state.

Please review the below video to see a basic overview.

Commits are fairly atomic and go into more detail as necessary.

https://github.com/bikebrigade/dispatch/assets/12987958/3d7a97a6-dd4b-4200-aa3e-14eed36f3c47

### DIFF
--- a/lib/bike_brigade_web/live/campaign_signup_live/index.ex
+++ b/lib/bike_brigade_web/live/campaign_signup_live/index.ex
@@ -99,7 +99,7 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Index do
   defp entity_in_campaigns?(socket, entity_campaign_id) do
     socket.assigns.campaigns
     |> Enum.flat_map(fn {_date, campaigns} -> campaigns end)
-    |> Enum.find(false, fn c -> c.id == entity_campaign_id end)
+    |> Enum.any?(fn c -> c.id == entity_campaign_id end)
   end
 
   attr :filled_tasks, :integer, required: true
@@ -134,7 +134,6 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Index do
     c = assigns.campaign
     filled_tasks = assigns.campaign_task_counts[c.id][:filled_tasks]
     total_tasks = assigns.campaign_task_counts[c.id][:total_tasks]
-    date_now = DateTime.utc_now()
     campaign_tasks_fully_assigned? = filled_tasks == total_tasks
     campaign_not_ready_for_signup? = is_nil(total_tasks)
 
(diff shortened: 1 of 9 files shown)


lib/bike_brigade_web/live/campaign_signup_live/index.ex
```diff
@@ -94,38 +94,100 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Index do
...
+    # Define map for button properties
+    button_properties = %{
+      in_past: {:disabled, "Completed"},
+      not_ready: {:disabled, "Campaign not ready for signup"},
+      fully_assigned: {:secondary, "Campaign Filled"},
+      rider_is_assigned: {:secondary, "Signed up for #{current_rider_task_count} deliveries"},
+      default: {:secondary, "Sign up"}
+    }
+
+    buttonType =
```
mveytsman, line 164: There's a bug here where I see a working signup button for past campaigns I've already signed up for

CleanShot 2024-03-12 at 14 59 44@2x

mveytsman: Also I think this map and the cond below it can be combined into one -- having the cond do the right rder (same as map?) I think will fix this bug

mveytsman (changes requested):
Couple of issues, mainly that we can signup for campaigns in the past (see below).


I found also if I navigate to a past campaign direclty I still see signup buttons

CleanShot 2024-03-12 at 15 01 27@2x

lib/bike_brigade_web/live/campaign_signup_live/index.ex
```diff
@@ -94,38 +94,100 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Index do
...
+      end
+
+    # Extract button properties based on buttonType
+    {button_color, signup_text} = Map.fetch!(button_properties, buttonType)
+
+    assigns =
+      assigns
+      |> assign(:signup_text, signup_text)
+      |> assign(:button_color, button_color)
+      |> assign(:c, c)
```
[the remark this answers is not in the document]
teesloane: Good catch, we do not.

lib/bike_brigade_web/live/campaign_signup_live/index.ex
teesloane: Good call, fixed that up.

lib/bike_brigade_web/live/campaign_signup_live/index.ex
```diff
@@ -145,46 +144,38 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Index do
         assigns.campaign_task_counts[c.id].rider_ids_counts[assigns.rider_id] || 0
       end
 
-    campaign_in_past =
-      case DateTime.compare(assigns.campaign.delivery_end, date_now) do
-        :gt -> false
-        :eq -> false
-        :lt -> true
-      end
+    campaign_in_past = campaign_in_past(assigns.campaign)
```
mveytsman, line 147: 👍

lib/bike_brigade_web/live/campaign_signup_live/show.ex
```diff
@@ -242,4 +256,9 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Show do
     </div>
     """
   end
+
+  defp task_eligible_for_signup(task, campaign) do
```
mveytsman, line 260: 👍

lib/bike_brigade_web/live/campaign_signup_live/show.html.heex
```diff
@@ -1,56 +1,74 @@
...
-      - <%= @campaign.program.campaign_blurb %>
-    </span>
+    <span class="text-sm font-bold">Pickup: <%= @campaign.location.address %></span>
+    <span :if={@campaign.program.campaign_blurb}>- <%= @campaign.program.campaign_blurb %></span>
   </:subtitle>
 </.header>
 
-<section>
-  <.table id="tasks" rows={@tasks}>
+<% sorted_tasks = Enum.sort_by(@tasks, & &1.id) %>
```
mveytsman, line 13: Is this just to make the tasks sorting consistent across page loads?

I think we can fix the query to sort the tasks inside of `campaign_riders_and_tasks`. WDYT?

mveytsman (changes requested):
Looks good!

One question in the code and I found you can still unassign yourself from campaigns in the past

CleanShot 2024-03-14 at 18 25 37@2x

We discussed a seperate ticket to make sure there's a window in which you can't unassign yourself, but I think before that we should disable that button for campaigns in the past,

teesloane:
> Looks good!
> 
> One question in the code and I found you can still unassign yourself from campaigns in the past
> 
> CleanShot 2024-03-14 at 18 25 37@2x
> We discussed a seperate ticket to make sure there's a window in which you can't unassign yourself, but I think before that we should disable that button for campaigns in the past,

Fixed:

https://github.com/bikebrigade/dispatch/assets/12987958/647e490a-5377-4798-999d-8f11cdba0f27

mveytsman (approved):
🎉 Amazing!

PR #292, opened 2024-03-14
Migrate: campaign_riders to have rider_signed_up.

teesloane:
## Why?

We want to be able to discern when a rider themself is responsible for signing up for a campaign versus a dispatcher signing up a rider. This way we'll be able to better create reports and see the effectiveness of the rider portal.

## Notes

I was trying to think of how I could test this on the frontend, but couldn't think of any ways that the ui changes as a result of this addition.  I could add tests to the `delivery.ex` context I suppose?


mveytsman (approved):
I think you can make a test that signs up a rider via the UI and makes sure that the field is set - this makes sense for a LiveView test but not for a fully frontend test.

I think its fine to ship this without that particular test though

PR #299, opened 2024-03-20
Fix overflowing tag list

mveytsman:
Go from 
CleanShot 2024-03-20 at 17 41 13@2x

to

CleanShot 2024-03-20 at 17 42 44@2x


teesloane (approved):
🚢 !

PR #305, opened 2024-03-26
Add show/hide button to the leaderboard next to your name

mveytsman:
Based on Darlene's feedback I added a button next to your name to show/hide yourself

CleanShot 2024-03-26 at 11 36 36@2x

CleanShot 2024-03-26 at 11 36 29@2x

### DIFF
--- a/lib/bike_brigade_web/live/stats_live/leaderboard.ex
+++ b/lib/bike_brigade_web/live/stats_live/leaderboard.ex
@@ -195,10 +195,27 @@ defmodule BikeBrigadeWeb.StatsLive.Leaderboard do
 
     class = if is_current_rider, do: "bg-yellow-200 p-1", else: "p-1"
 
-    assigns = assign(assigns, name: name, class: class)
+    assigns =
+      assign(assigns,
+        name: name,
+        class: class,
+        is_anonymous: is_anonymous,
+        is_current_rider: is_current_rider
+      )
 
     ~H"""
-    <span class={@class}><%= @name %></span>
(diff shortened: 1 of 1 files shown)


lib/bike_brigade_web/live/stats_live/leaderboard.ex
```diff
@@ -195,10 +195,27 @@ defmodule BikeBrigadeWeb.StatsLive.Leaderboard do
...
+        is_anonymous: is_anonymous,
+        is_current_rider: is_current_rider
+      )
 
     ~H"""
-    <span class={@class}><%= @name %></span>
+    <div class="flex flex-col sm:flex-row">
+      <span class={@class}><%= @name %></span>
+      <.button
+        :if={is_current_rider}
```
teesloane, line 210: This is missing an `@` resulting in a compiler warning:

```

warning: you are accessing the variable "is_current_rider" inside a LiveView template.

Using variables in HEEx templates are discouraged as they disable change tracking. You are only allowed to access variables defined by Elixir control-flow structures, such as if/case/for, or those defined by the special attributes :let/:if/:for. If you are shadowing a variable defined outside of the template using a control-flow structure, you must choose a unique variable name within the template.
```

teesloane (approved):
LGTM - I noted one compiler warning.

lib/bike_brigade_web/live/stats_live/leaderboard.ex
mveytsman: Good catch!

PR #313, opened 2024-04-10
Fix: hide assigned tasks if campaign in past on rider signup page.

### LINKED ISSUE #306: Don't show "N available" on campaign singup if campaign is in past
CleanShot 2024-03-29 at 21 10 08@2x

teesloane:
Fixes #306 . 

**Before/After:**

CleanShot 2024-04-10 at 14 33 25@2x

### DIFF
--- a/lib/bike_brigade_web/live/campaign_signup_live/index.ex
+++ b/lib/bike_brigade_web/live/campaign_signup_live/index.ex
@@ -127,7 +127,7 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Index do
           {"text-gray-600", "N/A"}
 
         campaign_in_past(assigns.campaign) ->
-          {"text-gray-600", "N/A"}
+          {"text-gray-600", "Campaign over"}
 
         assigns.total_tasks - assigns.filled_tasks == 0 ->
           {"text-gray-600", "Fully Assigned"}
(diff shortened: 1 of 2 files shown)


lib/bike_brigade_web/live/campaign_signup_live/index.ex
```diff
@@ -122,13 +122,18 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Index do
...
-        case assigns.total_tasks - assigns.filled_tasks do
-          0 -> {"text-gray-600", "Fully Assigned"}
-          _ -> {"text-red-400", "#{assigns.total_tasks - assigns.filled_tasks} Available"}
-        end
+      cond do
+        assigns.filled_tasks == nil ->
+          {"text-gray-600", "N/A"}
+
+        campaign_in_past(assigns.campaign) ->
+          {"text-gray-600", "N/A"}
```
mveytsman, line 130: I think instead of "N/A" this should say "Complete" 

WDYT?

mveytsman (changes requested):

lib/bike_brigade_web/live/campaign_signup_live/index.ex
mveytsman: Or  "Campaign over" to match https://github.com/bikebrigade/dispatch/pull/314

mveytsman (approved):
LGTM!

PR #314, opened 2024-04-12
Fix: Display "campaign over" if campaign in past

### LINKED ISSUE #312: Small UI fixes for Rider task view
CleanShot 2024-04-10 at 14 29 07@2x

- There is an empty space where the button would be (if the campaign is in the past)
- Neighbourhoods, while likely to span two lines, might as well be put on one line and wrap to the next line if they are long.

teesloane:
Fixes #312 

### Changes
- Screenshot demonstrates showing a task a rider can signup for in the present, then moving the campaign into the past, then showing that the "campaign is over".
- Also updates the address to be one line.

https://github.com/bikebrigade/dispatch/assets/12987958/82305681-61b5-4e76-94ba-6315a5a6bb60


mveytsman (approved):
Looks good!

PR #315, opened 2024-04-12
Enable login for all newly created riders

### LINKED ISSUE #308: Enable login for all riders
Switch all riders to be able to signup, and create user accounts for all riders.

This would involve calling `Accounts.create_user_for_rider` for every rider (this can be a one-off task) and calling that on rider creation.

mveytsman:
This PR let's new riders log in automatically after they are created

Closes #308 

For the existing riders I'm going to manually create accounts for them in the IEx shell

### DIFF
--- a/lib/bike_brigade/tasks/mailchimp_importer.ex
+++ b/lib/bike_brigade/tasks/mailchimp_importer.ex
@@ -1,6 +1,7 @@
 defmodule BikeBrigade.Tasks.MailchimpImporter do
   import Ecto.Query, warn: false
   import BikeBrigade.Utils, only: [get_config: 1, with_default: 2]
+  alias BikeBrigade.Accounts
   alias BikeBrigade.Repo
   alias BikeBrigade.Locations.Location
   alias BikeBrigade.Tasks.Importer
@@ -80,7 +81,7 @@ defmodule BikeBrigade.Tasks.MailchimpImporter do
         Riders.update_rider(rider, rider_attrs)
 
       nil ->
-        Riders.create_rider(rider_attrs)
+        Riders.create_rider_with_user(rider_attrs)
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/bike_brigade/tasks/mailchimp_importer.ex
```diff
@@ -80,7 +81,7 @@ defmodule BikeBrigade.Tasks.MailchimpImporter do
         Riders.update_rider(rider, rider_attrs)
 
       nil ->
-        Riders.create_rider(rider_attrs)
+        Riders.create_rider_with_user(rider_attrs)
```
teesloane, line 84: Ah, so this is how riders are created - from mailchimp?

lib/bike_brigade_web/live/rider_live/form_component.ex
```diff
@@ -154,6 +154,7 @@ defmodule BikeBrigadeWeb.RiderLive.FormComponent do
     save_rider_edit_impl(socket, rider_form_params)
   end
 
+  # TODO: this can probably be deleted: we don't really create riders from the form?
```
teesloane, line 157: Can we:
a) fix that in this PR
b) create a ticket for it?

I tried navigating to `/riders/new` and the view crashed because of a missing apply_action.

teesloane (approved):

lib/bike_brigade/tasks/mailchimp_importer.ex
mveytsman: Yup!

lib/bike_brigade_web/live/rider_live/form_component.ex
mveytsman, line 157:
```suggestion
  # TODO: this can probably be deleted: we don't really create riders from the form?
  # https://github.com/bikebrigade/dispatch/issues/321
```

PR #318, opened 2024-04-14
Fix: make dispatcher mcgee indeed a dispatcher.

teesloane:
In setting up the repo from fresh, I noted that on a clean DB, logging in tries to route you to /home - thinking that the user is a user. This is because the seeded dispatcher does not have is_dispatcher set to true.


mveytsman (approved):
👍

PR #322, opened 2024-04-15
Fix `handle_info` on campaign signup

mveytsman:
This will h andle messages we don't care about without crashing

Fixes: https://app.honeybadger.io/projects/79471/faults/106666591

### DIFF
--- a/lib/bike_brigade_web/live/campaign_signup_live/index.ex
+++ b/lib/bike_brigade_web/live/campaign_signup_live/index.ex
@@ -65,6 +65,10 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Index do
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/bike_brigade_web/live/campaign_signup_live/index.ex
```diff
@@ -137,7 +141,7 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Index do
       |> assign(:copy, copy)
 
     ~H"""
-    <p class="flex flex-col md:flex-row items-center mt-0 text-sm text-gray-700">
+    <p class="flex flex-col items-center mt-0 text-sm text-gray-700 md:flex-row">
```
mveytsman, line 144: Oops that's my class re-ordering plugin

PR #324, opened 2024-04-16
Add concept of `private` campaigns

### LINKED ISSUE #300: Suppress campaigns from rider facing interface
There are campaigns that are not open to all riders (Dixon Hall and West Neighbourhood house). Ideally, campaigns would have a checkbox to prevent them from displaying to all riders. 
image

mveytsman:
Added `public` flag to campaigns

It's defaulting to false for now as we are not opting in all campaigns for the portal just yet. Duplicating campaigns will duplicate its public/private state.

cc @sereprz - we should opt in the campaigns we're currently testing with, and make sure dispatchers know about this when you do the demo!

https://github.com/bikebrigade/dispatch/assets/34720/a6a42993-d177-4ed2-8341-6864a0b79d0a

Closes #300

### DIFF
--- a/lib/bike_brigade/delivery.ex
+++ b/lib/bike_brigade/delivery.ex
@@ -151,6 +151,9 @@ defmodule BikeBrigade.Delivery do
     # (such as when we need to display campaigns that urgently need a rider)
     campaign_ids = Keyword.get(opts, :campaign_ids, nil)
 
+    public = Keyword.get(opts, :public, nil)
+
+
     query =
       from c in Campaign,
         as: :campaign,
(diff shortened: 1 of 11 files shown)


teesloane (approved):
Tested and looks good and works great!

sereprz:
Noted! The flag can only be turned on by us am I right? I think we'll turn them on one-two at a time and I'm hoping to have a schedule for which campaigns go live when in the next week

mveytsman:
Any dispatcher can change the flag - and it carries in duplicating so if they follow the duplication workflow they will be able to carry forward the piblic/private status

mveytsman:
Closing in favour of #328

PR #327, opened 2024-04-17
Add Task Assignment Log

### LINKED ISSUE #323: Event log table for rider signups
In order to facilitate analytics about when riders are signing up (or cancelling) deliveries we need an event log table

The schema @sereprz and I came up with is 

```
task_id references(task)
rider_id references(rider)
user_id references(rider)
timestamp datetime
action string/enum
```

Actions so far are `assigned`, `unassigned`

mveytsman:
Closes #323 

This adds a `TaskAssignmentLog` which tracks `user_id`, `rider_id`, `timestamp`, and `action` (`:assigned`, or `:unassigned`) for all assignment (whether rider directly or dispatcher).

### DIFF
--- a/lib/bike_brigade/delivery.ex
+++ b/lib/bike_brigade/delivery.ex
@@ -7,6 +7,7 @@ defmodule BikeBrigade.Delivery do
   import Geo.PostGIS, only: [st_distance: 2]
   alias BikeBrigade.Riders.Rider
 
+  alias BikeBrigade.History
   alias BikeBrigade.Messaging
   alias BikeBrigade.Delivery.{Task, CampaignRider, TaskAssignmentLog}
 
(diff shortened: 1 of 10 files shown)


teesloane (approved):
Looks good. Added a small question.

lib/bike_brigade/delivery.ex
```diff
@@ -1012,4 +1056,19 @@ defmodule BikeBrigade.Delivery do
...
+  """
+  def list_task_assignment_logs() do
+    Repo.all(TaskAssignmentLog)
+  end
+
+  def create_task_assignment_log(attrs \\ %{}) do
+    %TaskAssignmentLog{timestamp: DateTime.utc_now()}
+    |> TaskAssignmentLog.changeset(attrs)
+    |> Repo.insert()
+  end
```
teesloane, line 1073: Nit: do you think this might belong in a new context, "tracking" or named something similar?

mveytsman: @teesloane what do you mean by on the main PR?

I actually think that putting this in a new context makes sense!

teesloane: When I wrote this comment, I wrote it while going through each commit one by one; I figured it might get lost when submitting my review; I meant to write the comment on the page for the PR as a whole, if that makes sense (typing this now, it doesn't make sense, but now typing that it doesn't make sense, — well, it's all starting to make sense.)

He typed from a chair as the wind battered at the trees outside his window. It was getting late and Ty was thinking about going to bed.
xo

PR #329, opened 2024-04-17
Only show first initial for recipient

mveytsman:
Our spreadsheets never show recipient names, just first initials.

Before 
CleanShot 2024-04-17 at 17 07 02@2x

After
CleanShot 2024-04-17 at 17 07 21@2x

### DIFF
--- a/test/bike_brigade_web/live/campaign_signup_live_test.exs
+++ b/test/bike_brigade_web/live/campaign_signup_live_test.exs
@@ -139,7 +139,11 @@ defmodule BikeBrigadeWeb.CampaignSignupLiveTest do
 
     test "we see pertinent task information", ctx do
       {:ok, _live, html} = live(ctx.conn, ~p"/campaigns/signup/#{ctx.campaign.id}/")
-      assert html =~ ctx.task.dropoff_name
+
+      # we only show the first initial of the dropoff name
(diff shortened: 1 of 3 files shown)


test/bike_brigade_web/live/campaign_signup_live_test.exs
```diff
@@ -139,7 +139,11 @@ defmodule BikeBrigadeWeb.CampaignSignupLiveTest do
 
     test "we see pertinent task information", ctx do
       {:ok, _live, html} = live(ctx.conn, ~p"/campaigns/signup/#{ctx.campaign.id}/")
-      assert html =~ ctx.task.dropoff_name
+
+      # we only show the first initial of the dropoff name
+      refute html =~ ctx.task.dropoff_name
+      assert html =~ ctx.task.dropoff_name |> String.first() |> String.upcase()
```
mveytsman, line 145: @teesloane this is searching the html for a letter, which is trivially always true. Any suggestion for how to make this actually test what I want?

teesloane: we could add unique id's to each task such that they show up as `unique-task-{task-id}-{first-letter}` then check that `has_element` is true?

PR #330, opened 2024-04-22
Show public campaigns as well as opportunities in the the calendar

mveytsman:
This updates the calendar at https://www.bikebrigade.ca/deliveries to show public Campaigns as well as Opportunities.

The idea is that dispatchers will no longer need to update the opportunities with signup links - any public campaigns will just show up on https://www.bikebrigade.ca/deliveries with the correct link within the portal.

We're demoing the portal to dispatchers on Tuesday so I'm pushing this now, but there are two outstanding tasks for later:

- [ ] Some programs we hide the pickup address for since it's a residential unit (Maggies is one I believe). We should move that flag from `Opportunities` to `Programs` so we can maintain that feature.
- [ ] Add some copy / prevent dispatchers from updating opportunities for campaigns that are launched with the portal (e.g. the program is public)


teesloane (approved):
Tested and was able to see a public program show it's campaign info in the embedded calendar!

CleanShot 2024-04-22 at 16 35 57@2x

I did try creating an opportunity in order to see that show up in the embed calendar but for some reason I wasn't able to create one through the form. Might just want to double check that you can do that from your local and that I'm not missing anything.

mveytsman:
@teesloane I fixed this issue (looks like we had it in prod too 🙃 ) and did the testing you were trying to do!

mveytsman:
I added some alerting when editing opportunities:

When you go to the opportunity page:
CleanShot 2024-04-22 at 19 46 33@2x


If you try to edit an opportunity for a public campaign:

CleanShot 2024-04-22 at 19 46 29@2x

sereprz:
This is great! Small suggestion - I would replace `Alert` with `Warning` if it takes less than 5 min

mveytsman:
How did you know that this is in 20 different files 😁 https://github.com/bikebrigade/dispatch/pull/330/commits/67c73ac6e7e29c98b6c1b181bf960a0f74b65561

anyways will do :)

PR #343, opened 2024-04-26
Fix: sort tasks by neighbourhoods.

teesloane:
While pairing with @chadmohr he mentioned that we might want to consider grouping tasks by neighbourhood — sometimes tasks are all going to the same building / is useful for riders who can/want to take multiple deliveries to the same location. 

### Before

CleanShot 2024-04-26 at 11 26 08@2x

### After

CleanShot 2024-04-26 at 11 25 12@2x

### DIFF
--- a/lib/bike_brigade/delivery.ex
+++ b/lib/bike_brigade/delivery.ex
@@ -458,9 +458,7 @@ defmodule BikeBrigade.Delivery do
       |> Repo.preload(:location)
 
     # Does a nested preload to get tasks' assigned riders without doing an extra db query
-    tasks =
-      Repo.preload(all_tasks, assigned_rider: fn _ -> all_riders end)
-    |> Enum.sort_by(fn t -> t.dropoff_location.neighborhood.name end)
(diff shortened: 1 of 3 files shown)


lib/bike_brigade/delivery.ex
```diff
@@ -458,7 +458,9 @@ defmodule BikeBrigade.Delivery do
       |> Repo.preload(:location)
 
     # Does a nested preload to get tasks' assigned riders without doing an extra db query
-    tasks = Repo.preload(all_tasks, assigned_rider: fn _ -> all_riders end)
+    tasks =
+      Repo.preload(all_tasks, assigned_rider: fn _ -> all_riders end)
+    |> Enum.sort_by(fn t -> t.dropoff_location.neighborhood.name end)
```
teesloane, line 463: tried to figure out how to do this at the DB level for too long and gave up. Happy to 🍐 if you know a better solution.

mveytsman (approved):
LGTM!

PR #344, opened 2024-04-26
Show initials for riders including last name

### LINKED ISSUE #337: Recipient initials
Is it possible to get the first letter and then any additional letters following a space from the name field for recipients on the rider facing signups? Looking at something like the attached, with one letter and without the crossroads, it takes me a surprising amount of time to figure out who is who, despite having worked with this list every week for over 2 years and having met all of these individuals :P

image

mveytsman:
Closes #337


teesloane (approved):
🚀 !

PR #348, opened 2024-04-27
Fix: crashing sort by on neighbourhoods

teesloane:
Currently, the signup show.ex file crashes if a task doesn't have a neighbourhood - `dropoff_location.neighborhood` can be nil - this handles for that case.

### DIFF
--- a/lib/bike_brigade_web/live/campaign_signup_live/show.ex
+++ b/lib/bike_brigade_web/live/campaign_signup_live/show.ex
@@ -172,7 +172,12 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Show do
 
   defp assign_campaign(socket, campaign) do
     {riders, tasks} = Delivery.campaign_riders_and_tasks(campaign)
-    tasks = Enum.sort_by(tasks, fn t -> t.dropoff_location.neighborhood.name end)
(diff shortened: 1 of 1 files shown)


lib/bike_brigade_web/live/campaign_signup_live/show.ex
```diff
@@ -172,7 +172,12 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Show do
 
   defp assign_campaign(socket, campaign) do
     {riders, tasks} = Delivery.campaign_riders_and_tasks(campaign)
-    tasks = Enum.sort_by(tasks, fn t -> t.dropoff_location.neighborhood.name end)
+    tasks = Enum.sort_by(tasks, fn t ->
```
mveytsman, line 175: Why not use `Location.neighborhood` here which handles nils?

mveytsman (changes requested):
Good catch!

Minor suggesiton inline

lib/bike_brigade_web/live/campaign_signup_live/show.ex
teesloane: Ooh, I forgot.

mveytsman (approved):
🎉

Issue #340, opened 2024-04-25
Displaying notes on signups - possible privacy concerns 


chadmohr:
There are cases where names, contact info and even other personal info is put in notes. Basically, there could be anything in the notes. And these notes are displayed on desktops in the rider signups.
 
![Untitled](https://github.com/bikebrigade/dispatch/assets/34720/6e6dc646-38fb-4944-8004-893c29bdc98a)

mveytsman:
This is a good catch!

Based on the convo in slack we'll hide these before signup. Cecily also suggested splitting up rider notes and signup notes in the future

@chadmohr I went ahead and edited the image to block the phone number, do you mind doing that in the future if you're uploading screenshots with private info?

chadmohr:
Yes, for sure. This occurred to me after I posted.

chadmohr:
Also, somewhere in here I am 99% sure I also suggested a signup notes box in addition to the delivery notes

sereprz:
We're going to hide the column for now and revisit how to handle notes later on e.g. notes on signup VS notes on delivery

chadmohr:
349 is closed. 

I don’t really understand the hesitation with the delivery and signup notes. Seems to me adding the latter solves a lot of problems.

chadmohr:
@teesloane the notes bstill show on mobile 
![image](https://github.com/bikebrigade/dispatch/assets/1443657/bc0ac3b2-822c-4a36-87d9-2d0a486e37a7)

PR #349, opened 2024-04-28
Fix: swap pickup notes for rider_notes on campaign singup.

teesloane:
Fixes #340

### DIFF
--- a/lib/bike_brigade_web/live/campaign_signup_live/show.html.heex
+++ b/lib/bike_brigade_web/live/campaign_signup_live/show.html.heex
@@ -23,8 +23,8 @@
     <:col :let={task} label="Dropoff Neighbourhood">
       <%= Locations.neighborhood(task.dropoff_location) %>
     </:col>
-    <:col :let={task} label="Notes"><%= task.delivery_status_notes %>
-      <.truncated_riders_notes note={task.rider_notes || "--"} /></:col>
+    <:col :let={task} label="Pickup Notes">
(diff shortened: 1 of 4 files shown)


lib/bike_brigade_web/live/campaign_signup_live/show.html.heex
```diff
@@ -23,8 +23,8 @@
     <:col :let={task} label="Dropoff Neighbourhood">
       <%= Locations.neighborhood(task.dropoff_location) %>
     </:col>
-    <:col :let={task} label="Notes"><%= task.delivery_status_notes %>
```
teesloane, line 26: While swapping this, I noticed that for some reasons I added the delivery_status_notes into this html table. 🧹

mveytsman:
Just to be clear we haven't added any form to fill them in yet right?

teesloane:
> Just to be clear we haven't added any form to fill them in yet right?

I did actually add it to the task form. I should have added a screenshot:

CleanShot 2024-04-28 at 16 02 45@2x

mveytsman:
Oh sorry, I missed the html bits. 

To be honest, I'm not 100% sure on this: I think the issue that this has PII and should be hidden from the public page should be addressed.

I think before adding a separate field we should iterate on clear instructions and train dispatchers e.g. I notice there's a todo to rename the field from rider notes to delivery instructions.

What if we make a PR to hide it and then get some feedbakc on slack on copy for adding a second field?

teesloane:
Sure, I'll make a PR to hide the delivery notes for now and then we can circle back to this when necessary :)

Issue #352, opened 2024-04-29
Text riders for campaigns that are switching on the Sunday


mveytsman:
Schedule for switching here: https://docs.google.com/spreadsheets/d/1U587z_EOU4m4WQ1kdTGBt8RKiqMXMBis94nx-KsX3XU/edit#gid=1462122027

Text the riders that have ridden in the last 90 days for the given program on the sunday that they switch.

Copy of text message to follow.

sereprz:
Here's a template from Nathalie:

```
Howdy & heads up! Your Bike Bridage campaign is moving to the (awesome new!) Rider Portal this week. (Bye bye Google Sheets!) If you have any Qs or need any help, see: https://www.bikebrigade.ca/portal
```

My personal take is that we replace `Your Bike Brigade campaign` with the name of the campaign IF it's a quick thing that we can template. This is sending to the tutorial and I think that's okay? Or we could change it to `login using your phone number at [dispatch] or see check out a short tutorial at [https://www.bikebrigade.ca/portal]` or something to that effect

mveytsman:
Set up all the messages as campaigns - added links to the spreadsheet here: https://docs.google.com/spreadsheets/d/1U587z_EOU4m4WQ1kdTGBt8RKiqMXMBis94nx-KsX3XU/edit#gid=1462122027

PR #354, opened 2024-04-30
Message riders even if they are not assigned any tasks

mveytsman:
There's a little-known "feature" where we only message riders on campaigns if they're assigned a delivery.

I want to get rid of it for three reasons

1) It's counterintuitive and not documented well for dispatchers
2) As we roll out the rider portal, dispatchers aren't going to be adding riders to campaigns so it's not pertinent anymore
3) I want to use the campaigns' scheduled messanger feature for a hacky way to do scheduled bulk messages announcing the portal (#352 ) and this has to happen for it to work

@chadmohr I think you might be the only person that's aware of this "feature" - pinging you in case you have any concerns

### DIFF
--- a/lib/bike_brigade/delivery.ex
+++ b/lib/bike_brigade/delivery.ex
@@ -572,7 +572,7 @@ defmodule BikeBrigade.Delivery do
     {riders, _} = campaign_riders_and_tasks(campaign)
 
     msgs =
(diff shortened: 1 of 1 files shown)


teesloane (approved):

chadmohr:
I have not come across any instances of bulk messaging for a long time other than what I w done

Issue #370, opened 2024-05-22
Delivery notes showing on mobile signups


chadmohr:
These were removed from desktop display due to privacy concerns, but still appear in mobile 
![image](https://github.com/bikebrigade/dispatch/assets/1443657/cc77ed40-2c54-4301-833e-978ec22ff292)

mveytsman:
I'm addressing this as part of #364

mveytsman:
Closed in #372

PR #372, opened 2024-05-22
Show item name and description in signup

### LINKED ISSUE #364: Change delivery signup to show the name of the item being delivered (instead of the category)
[This code](https://github.com/bikebrigade/dispatch/blob/c882145a6721036751137a6dcafbbda8f0a99fb0/lib/bike_brigade_web/live/campaign_signup_live/show.ex#L185-L199) should have the `category` removed and replaced with the `name`. 

edit: we should test this with multiple items and make sure they all show up :)

mveytsman:
We show the item name and description when signing up now. I used the description/summary thing with minimal styling - but honestly I think it's good for now.

Note also I got rid of the number when it's one item

## Desktop
CleanShot 2024-05-22 at 16 36 34

## Mobile
CleanShot 2024-05-22 at 16 35 21

Closes #364 #370

### DIFF
--- a/lib/bike_brigade_web/live/campaign_signup_live/show.ex
+++ b/lib/bike_brigade_web/live/campaign_signup_live/show.ex
@@ -183,16 +183,10 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Show do
   ## Module specific components
 
   defp get_delivery_size(assigns) do
-    item_list =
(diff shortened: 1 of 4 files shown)


teesloane (approved):
Nice. You could also add `title` attributes to the summary so that on mouse over we get alt text showing. But I think this is good to go now.

PR #375, opened 2024-05-23
Add prompt if the rider is unassigning a delivery today

### LINKED ISSUE #294: Don't allow riders to unassign themselves within N hours of the delivery
Right now you could unassign yourself from a task a minute before a delivery is scheduled. This will be an issue for dispatchers as people may just click that button last minute.

Within a certain time-period of the delivery start (to be determined in conversation with staff), riders should not be able to unassign themselves from the delivery. Instead they should be told to contact dispatch (currently via text message) to do so.

mveytsman:
This uses the default confirm dialog to add a prompt if the user is cancelling a delivery today. I didn't add the prompt for texting since it's easier to do this as a first pass.

Closes #294 (for now)

## Desktop 
CleanShot 2024-05-23 at 18 28 33@2x

## Mobile
image


teesloane (approved):
Looks good.

What do you thin about including the number to test in the alert copy?

PR #378, opened 2024-05-27
Show "You" when assigned

### LINKED ISSUE #360: Provide clearer success state / copy  on rider signup
Some riders are expressing confusion when they have signed up for a delivery:

- When the "sign up" button for a delivery turns to a red "unassigne me" button, some people are confused.
- On clicking signup we should show some kind of "success" state - either a flash or a modal that pops up for however many seconds and says `you signed up to deliver X on Y date! you will recieve a notification before your delivery with further instructions`
- instead of showing "unassign me" we could should the person's name? Or maybe their name and "cancel delivery" next to it? @chadmohr I think you might have better ideas.

mveytsman:
Show "You" when someone is the person assigned

## Desktop
CleanShot 2024-05-27 at 13 38 21

## Mobile
CleanShot 2024-05-27 at 13 42 45

I think this is enough and we don't need a flash necessarily.

Closes #360

### DIFF
--- a/lib/bike_brigade_web/live/campaign_signup_live/show.ex
+++ b/lib/bike_brigade_web/live/campaign_signup_live/show.ex
@@ -215,11 +215,15 @@ defmodule BikeBrigadeWeb.CampaignSignupLive.Show do
 
   def signup_button(assigns) do
     ~H"""
-    <div>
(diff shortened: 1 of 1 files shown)


chadmohr:
Love it

teesloane (approved):

PR #382, opened 2024-06-03
fix: reposition flash on mobile.

### LINKED ISSUE #346: Flash on mobile obscures menu bar until closed
CleanShot 2024-04-26 at 13 07 56@2x

We should use some JS to dismiss flashes after a certain amount of time.

teesloane:
## Describe your changes

closes #346

I was originally overthinking this - adding a new flash library is fairly involved. Why not just move the flash down to within thumbs reach on mobile (and get it out of the way of menu).

If we need to explore future flash libraries, [Flashy](https://github.com/sezaru/flashy) looks good, and so does [LiveToast](https://github.com/srcrip/live_toast). Both involve having to setup JS hooks. 

## Screenshot

https://github.com/bikebrigade/dispatch/assets/12987958/3a354b4c-c102-4d91-ae2c-72bde00e593f

- [n/a] If it is a core feature, I have added tests.
- [n/a] Are there other PRs or Issues that I should link to here?
- [n/a] Will this be part of a product update? If yes, please write one phrase

### DIFF
--- a/lib/bike_brigade_web/core_components.ex
+++ b/lib/bike_brigade_web/core_components.ex
@@ -212,8 +212,8 @@ defmodule BikeBrigadeWeb.CoreComponents do
       phx-mounted={@autoshow && show("##{@id}")}
       phx-click={JS.push("lv:clear-flash", value: %{key: @kind}) |> hide("##{@id}")}
       class={[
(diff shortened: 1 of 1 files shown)


mveytsman (approved):
LGTM

And thanks for adding libraries to look into

PR #383, opened 2024-06-06
Add: "N available" copy to the dispatcher side.

teesloane:
## Describe your changes

Up for discussion: adding clarification copy about remaining slots to sign up for on the dispatcher side. Originally suggested  up by @chadmohr in this [slack thread](https://thebikebrigade.slack.com/archives/C016VGHETS4/p1717624178300409).

## Screenshot

https://github.com/bikebrigade/dispatch/assets/12987958/de19449a-440e-4eea-90a8-6a11f4f4b5f8

- [X] I have performed a self-review of my code

### DIFF
--- a/lib/bike_brigade_web/live/campaign_live/index.ex
+++ b/lib/bike_brigade_web/live/campaign_live/index.ex
@@ -7,11 +7,16 @@ defmodule BikeBrigadeWeb.CampaignLive.Index do
   alias BikeBrigade.Delivery
   alias BikeBrigade.Delivery.Campaign
   alias BikeBrigade.Messaging.SmsMessage
+  alias BikeBrigadeWeb.Components.CampaignComponents
 
   import BikeBrigadeWeb.CampaignHelpers
 
   @impl true
   def mount(_params, _session, socket) do
+    if connected?(socket) do
+      Delivery.subscribe()
+    end
+
     current_week =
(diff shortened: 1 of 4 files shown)


lib/bike_brigade_web/live/campaign_live/index.ex
```diff
@@ -21,14 +26,33 @@ defmodule BikeBrigadeWeb.CampaignLive.Index do
...
+
+  @impl Phoenix.LiveView
+  def handle_info({event, entity}, socket) when event in @broadcasted_infos do
+    if entity_in_campaigns?(socket.assigns.campaigns, entity.campaign_id) do
+      {:noreply, refetch_and_assign_data(socket)}
+    else
+      {:noreply, socket}
+    end
+  end
+
```
mveytsman, line 55: Thanks for adding this!

two feedbacks:

1) I think this should go after the handle_events
2) It needs a 
```
@impl Phoenix.LiveView
  @doc "silently ignore other kinds of messages"
  def handle_info(_, socket), do: {:noreply, socket}
```
after it to silently ignore messages we don't care about (or else the LV crashes :()

mveytsman (changes requested):
Looks good except small bit of feedback!

teesloane:
Ready for re-review @mveytsman !

PR #385, opened 2024-06-07
Add: signup_notes, rename rider_notes.

teesloane:
## Describe your changes

Follow up to the closing of #340 , closes #385

This was done so that we could include signup related information when riders signup for a delivery.
It has also been on the backlog to rename "rider_notes" to "delivery_instructions"

**Delivery Instructions** - To be shown when a rider is doing a delivery. Sometimes contains personal information (ex: "Francis is hearing impaired and might not hear when you ring the doorbell."

**Signup Notes** - Instructions specific to choosing what to sign up for - for example: "{recipient A} is in the same building as {recipient B} - if you can signup for both that would be awesome!

## Screenshot

https://github.com/bikebrigade/dispatch/assets/12987958/8a79cc64-4015-48c7-a0ca-a5708096c932

## Product Update

"Hi Dispatchers! It is now possible to include "rider signup notes—these will show on the page where a rider assigns themselves a specific delivery. This might be useful for letting riders know about co-located deliveries, for example. Please avoid adding personal identifying information into the rider signup notes."

- [x] Are there other PRs or Issues that I should link to here?
- [x] Will this be part of a product update? If yes, please write one phrase

### DIFF
--- a/lib/bike_brigade/delivery.ex
+++ b/lib/bike_brigade/delivery.ex
@@ -638,7 +638,7 @@ defmodule BikeBrigade.Delivery do
 
     task_details =
       for task <- tasks do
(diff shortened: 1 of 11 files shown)


mveytsman (approved):
Looks great!

PR #389, opened 2024-06-17
Convert the login page to use controllers

### LINKED ISSUE #307: Improve login expeirence on mobile
As reported by rider Jaime, and a few other people we have an issue with logging in from mobile where if you leave the page and come back during login it reloads so you go 

1) Enter phone number
2) Get text
3) Switch to messages app to get code from text
4) Switch to browser 
5) page reloads and you're asked to enter your phone number again

There are two things to do here:

- [x] Make it so the code from bike brigade is automatically detected by the phone and pastable (I don't know how to do this)
- [ ] Make it so the page doesn't boot you from entering your phone number when it reloads.

mveytsman:
## Describe your changes

Closes #307 

Switching to a controller / DeadView is enough and we don't need to do any cookie magic!

@teesloane I did a bit of refactoring after we paired so feel free to take another 👁️. Also #388  was needed for the `method="delete"` bit to work right. Glad I finally figured that one out after years of being broken 🙈 

- [x] If it is a core feature, I have added tests.
- [x] Are there other PRs or Issues that I should link to here?
- [x] Will this be part of a product update? If yes, please write one phrase


teesloane (approved):
synchronous review with @mveytsman. LGTM!

PR #390, opened 2024-06-25
Fix: missing function clause.

### LINKED ISSUE #355: [Bike Brigade/prod] FunctionClauseError: no function clause matching in BikeBrigadeWeb.CampaignLive.FormComponent.save_campaign/3
This happens when creating a new campaign from the program page, i.e. I made a campaign from clicking "new campaign" on https://dispatch.bikebrigade.ca/programs/79
## Backtrace

    line 116 of lib/bike_brigade_web/live/campaign_live/form_component.ex: save_campaign/3

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/79471/faults/86456460)

teesloane:
## Describe your changes

Closes #355.

Technically, I suppose I could rename :new_campaign over on the program show.ex file, but I wasn't sure if it was there for a reason (as perhaps having just `:new` might imply it's for making a `:new` program ? 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,8 @@ jobs:
       - name: Set up Elixir
         uses: erlef/setup-elixir@v1
         with:
-          elixir-version: '1.14.2'
-          otp-version: '25'
+          elixir-version: '1.15'
+          otp-version: '26'
 
       - name: Cache Elixir deps
         id: elixir-deps-cache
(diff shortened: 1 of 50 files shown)


mveytsman (changes requested):
This works but I think it's messy to bring in the `new_campaign` action from the program into the campaign form component

better would be to change the name of the action to `:new` when we load this component here: https://github.com/bikebrigade/dispatch/blob/main/lib/bike_brigade_web/live/program_live/show.html.heex#L87

I think that's the better separation of concerns

mveytsman (approved):

Issue #381, opened 2024-05-30
Go from one dispatch instance to two (set up for BikeBrigade Halifax)


mveytsman:
This is an epic to track the work that needs to be done to be able to launch an instance of bike brigade dispatch for another community. 

- [ ] Runbooks for launching (on fly.io for now) / all auxilary services you need to set up (twilio, database, mapbox, etc)
- [ ] Sign-ups are handled via mailchimp right now - the web form creates a new mailchimp contact, which creates a user via webhook. It may be necessary to have some other 
- [ ] `LocalizedDatatime` assumes EST. this timezone should be configarable per instance
- [ ] Neighborhoods (and [soon possibly intersections](#317)) are pulled from Toronto specific open data. Find some way to get other cities data in there or do something else to solve the problem of telling riders where things are before we reveal the exact address.

danp:
Re neighborhoods, a quick search indicates it's using some geojson data currently?

I happen to have [this](https://github.com/danp/outages-to-sqlite/tree/main/places) which I'm using to turn power outage locations into neighborhoods for HRM. So hopefully we can come up with a way to make the neighborhood geojson data switchable to use something like that!

PR #393, opened 2024-07-01
config: support APP_TZ to configure time zone

danp:
## Describe your changes

Currently, the app has "America/Toronto" hardcoded as the time zone in LocalizedDateTime. That makes it harder to use in other zones.

To help that, support APP_TZ in the environment. If set, use that value instead. The default is still America/Toronto.

Updates #381

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -9,6 +9,9 @@ import Config
 
 config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
 
+config :bike_brigade, BikeBrigade.LocalizedDateTime,
+  timezone: System.get_env("APP_TZ", "America/Toronto")
+
 config :bike_brigade,
   ecto_repos: [BikeBrigade.Repo]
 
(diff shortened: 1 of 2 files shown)


config/config.exs
```diff
@@ -9,6 +9,9 @@ import Config
 
 config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
 
+config :bike_brigade, BikeBrigade.LocalizedDateTime,
+  timezone: System.get_env("APP_TZ") || "America/Toronto"
```
mveytsman, line 13:
```suggestion
  timezone: System.get_env("APP_TZ", "America/Toronto")
```

Worth nothing that `System.get_env` takes a second optional argument for default

mveytsman (approved):
This is great!

And also I really didn't realize it would be so simple -- I assumed I had Toronto hardcoded everywhere. 

One minor nit but otherwise 🚀

PR #395, opened 2024-07-01
.github: bump CI elixir and OTP versions

danp:
## Describe your changes

Should fix these CI warnings:

the underscored variable "_t" is used after being set

Ref elixir-lang/elixir#12262


mveytsman (approved):
Good catch!!!

Issue #394, opened 2024-07-01
Format Elixir source with `mix format`?


danp:
As a newcomer both to Elixir and this app, it would be very helpful to not have to worry too much about formatting.

I'm still getting my editor/etc setup going which should probably auto-format on save.

But it would be nice if `mix format` use was encouraged/enforced to ensure some baseline consistency.

For example running `mix format mix.exs "lib/**/*.{ex,exs}" "test/**/*.{ex,exs}"` as described in `mix help format` makes quite a few changes.

mveytsman:
Good call. I have had mix format running on save before but have not consistently ran it as part of linting. 

I ran it in #396. 

Do you feel like we should add format as part of linting / a precommit hook?

danp:
👍 I think enforcing it as part of CI would be great, along with tips in README/etc to suggest how to format locally before pushing.

PR #404, opened 2024-07-04
.github,etc: enforce `mix format` in CI

danp:
## Describe your changes

Consistent and automated code formatting can make it much easier to work with a codebase.

In #396 a run of `mix format` with args was done.

This change enforces use of `mix format` in CI and adds a note to the README about how to format. A hint for editor and git config is also added.

The existing formatter config is fixed up so just running `mix format` should do the right thing.

Updates #394

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,8 @@ jobs:
         run: mix deps.get
       - name: Compile
         run: mix compile
+      - name: Check Elixir code formatting
+        run: mix format --check-formatted
       - name: Install JS dependencies
         run: npm install --prefix=assets
       - name: Compile assets
--- a/config/config.exs
+++ b/config/config.exs
@@ -68,10 +68,9 @@ config :phoenix, :json_library, Jason
 config :geo_postgis,
   json_library: Jason
 
-config :bike_brigade, BikeBrigade.Tasks.MailchimpImporter,
-  list_id: {:system, "MAILCHIMP_LIST_ID"}
+config :bike_brigade, BikeBrigade.Tasks.MailchimpImporter, list_id: {:system, "MAILCHIMP_LIST_ID"}
 
-  config :bike_brigade, BikeBrigade.Tasks.MailchimpAttributesSync,
+config :bike_brigade, BikeBrigade.Tasks.MailchimpAttributesSync,
   list_id: {:system, "MAILCHIMP_LIST_ID"}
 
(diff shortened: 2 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -58,6 +58,8 @@ jobs:
         run: mix deps.get
       - name: Compile
         run: mix compile
+      - name: Check Elixir code formatting
+        run: mix format --check-formatted
```
danp, line 62: Found this in https://fly.io/phoenix-files/github-actions-for-elixir-ci/.

config/config.exs
```diff
@@ -68,10 +68,9 @@ config :phoenix, :json_library, Jason
 config :geo_postgis,
   json_library: Jason
 
-config :bike_brigade, BikeBrigade.Tasks.MailchimpImporter,
-  list_id: {:system, "MAILCHIMP_LIST_ID"}
+config :bike_brigade, BikeBrigade.Tasks.MailchimpImporter, list_id: {:system, "MAILCHIMP_LIST_ID"}
```
danp, line 71: I think some of these changes are from the command run for #396 maybe not including the `config` dir? But not quite sure why stuff in `lib` is changed so much.

priv/repo/.formatter.exs
```diff
@@ -1,4 +1,4 @@
 [
   import_deps: [:ecto_sql],
-  inputs: ["/**/*.{ex,exs}"],
```
danp, line 3: When I first ran `mix format` it just hung and I wasn't sure why. Once I learned about `.formatter.exs` and `subdirectories` (which the top-level file uses) I found this extra leading `/`. With that I'm pretty sure it was considering all files on my system for formatting.

config/config.exs
mveytsman: I think the stuff in lib changed because mix format ran without the `:ecto` import dep - which we have in `.formatter.exs` in the main dir -- i wonder if you ran mix format from within the subdirectory and the bug you found below triggered.

Do you mind trying to undo all your changes in lib and running the formatter again - on my system it doens't add all the parenthesis which is what I want.

The idea is that you generally parenthesize, but the ecto dsl skips them which is what I'm hoping the formatter to enforce.

danp: 👍 that seems to have removed almost all changes from `lib`, thanks!

mveytsman (approved):
Ooops forgot to re-review these!

Looks great @danp

PR #429, opened 2025-03-27
Write some initial e2e tests

teesloane:
## Describe your changes

This PR adds Playwright e2e testing and a handful of example tests. A few notes:

- Notes on running the e2e test have been added to the readme
- A few small changes to the application code were required to get tests to be able to target dom elements
- It is about 90% reliable, and I haven't quite figured out why 10% of the time or tests seem to fail (or rather, they fail, and then when run again they pass). 
  - Depending on your internet  connection, you could run into issues with the Google maps auto complete working when creating a campaign
  - Sometimes creating a program just... fails when trying to click a text box

- [-] If it is a core feature, I have added tests.
- [-] Are there other PRs or Issues that I should link to here?
- [-] Will this be part of a product update? If yes, please write one phrase

### DIFF
--- a/lib/bike_brigade_web/live/campaign_live/form_component.ex
+++ b/lib/bike_brigade_web/live/campaign_live/form_component.ex
@@ -72,7 +72,7 @@ defmodule BikeBrigadeWeb.CampaignLive.FormComponent do
 
   @impl true
   def mount(socket) do
-    programs = for p <- Delivery.list_programs(), do: {p.name, p.id}
+    programs = for p <- Delivery.list_programs(), do: [key: p.name, value: p.id, label: p.name]
 
     {:ok,
      socket
--- a/mix.exs
+++ b/mix.exs
@@ -122,7 +122,8 @@ defmodule BikeBrigade.MixProject do
       "ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],
       "ecto.reset": ["ecto.drop", "ecto.setup"],
       test: ["ecto.create --quiet", "ecto.migrate", "test"],
-      "assets.deploy": ["esbuild default --minify", "tailwind default --minify", "phx.digest"]
+      "assets.deploy": ["esbuild default --minify", "tailwind default --minify", "phx.digest"],
+      "test.e2e": ["ecto.reset", "phx.server"]
     ]
   end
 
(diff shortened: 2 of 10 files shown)


lib/bike_brigade_web/live/campaign_live/form_component.ex
```diff
@@ -72,7 +72,7 @@ defmodule BikeBrigadeWeb.CampaignLive.FormComponent do
 
   @impl true
   def mount(socket) do
-    programs = for p <- Delivery.list_programs(), do: {p.name, p.id}
+    programs = for p <- Delivery.list_programs(), do: [key: p.name, value: p.id, label: p.name]
```
teesloane, line 75: This was required in order to add "label" attributes to options in the `<select>` tag; these labels are used to traget dropdowns in tests.

mix.exs
```diff
@@ -122,7 +122,9 @@ defmodule BikeBrigade.MixProject do
       "ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],
       "ecto.reset": ["ecto.drop", "ecto.setup"],
       test: ["ecto.create --quiet", "ecto.migrate", "test"],
-      "assets.deploy": ["esbuild default --minify", "tailwind default --minify", "phx.digest"]
+      "assets.deploy": ["esbuild default --minify", "tailwind default --minify", "phx.digest"],
+      # REVIEW: Sometimes this hangs on a query - `INSERT INTO "tasks_items"...`
+      "test.e2e": ["ecto.drop", "phx.server"]
```
teesloane, line 127: Not sure what's going on here—but it just seems to hang. Maybe just remove it?

test/e2e/playwright.config.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    {
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
+    },
+
+    {
+      name: 'webkit',
+      use: { ...devices['Desktop Safari'] },
+    },
```
teesloane, line 50: It might be worth removing testing for Firefox and Safari?

test/e2e/all.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  await page.getByRole('textbox', { name: 'Contact Name' }).press('Tab');
+  await page.getByRole('textbox', { name: 'Contact Email' }).fill('joecool@gmail.com');
+  await page.getByRole('textbox', { name: 'Contact Email' }).press('Tab');
+  await page.getByRole('textbox', { name: 'Contact Phone' }).fill('6475555554');
+  await page.getByRole('button', { name: 'Save' }).click();
+
+
+  // once the program is created, we need to find it on the program page and EDIT it,
+  // because at this time there is no "add items" for part when creating a new program
+  // (only when editing it)
```
teesloane, line 148: I wonder if this is an intentional part of the New Program form, or if the ability to add items should be possible in that form.

mveytsman:
Hmm is it possible to get this to at least run in the test database instead of dev in order to not wipe it?

PR #431, opened 2025-06-17
Enable de-anonymizing leaderboard stats when user is a dispatcher.

### LINKED ISSUE #430: Leaderboard shouldn't be anonymized for dispatcher
The leaderboard should show dispatchers the names of riders even when anonymized publicly.

It should still indicate who is anonymized for milestone purposes

teesloane:
## Describe your changes

Closes #430 

CleanShot 2025-06-17 at 08 47 43

- [x] If it is a core feature, I have added tests.

### DIFF
--- a/lib/bike_brigade_web/live/stats_live/leaderboard.ex
+++ b/lib/bike_brigade_web/live/stats_live/leaderboard.ex
@@ -117,7 +117,8 @@ defmodule BikeBrigadeWeb.StatsLive.Leaderboard do
   @impl Phoenix.LiveView
   def handle_event("toggle_all_riders_anonymity", _params, socket) do
     if socket.assigns.current_user.is_dispatcher do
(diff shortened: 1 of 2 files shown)


mveytsman (changes requested):
Looks good so far but missing a backend security check - right now we just hide the button for non-dispatchers but the event can still be sent.

Also, and maybe this is scope creep, do we allow non-dispatchers to download the full leaderboard as CSV which doesn't anonymize?

teesloane:
> Looks good so far but missing a backend security check - right now we just hide the button for non-dispatchers but the event can still be sent.

I added some server-side checking.

> Also, and maybe this is scope creep, do we allow non-dispatchers to download the full leaderboard as CSV which doesn't anonymize?

This is the logged-in-rider view - they cannot download a CSV unless they are a dispatcher.

CleanShot 2025-06-17 at 12 28 26@2x

mveytsman:
Looks great and I love the test, thanks Ty!

PR #440, opened 2025-08-09
Add live views for banner feature.

teesloane:
## Describe your changes

https://github.com/user-attachments/assets/70f17471-fd9b-4f24-b77e-701bf1796f6d

⏺ 📢 Add Banner System with Real-time Updates

  Features

  - Banner management interface - Admin can create/edit/delete scheduled banners with date/time pickers
  - Time-based display logic - Only shows banners that are currently active (between start/end times and enabled)
  - Validation prevents invalid schedules - Can't create banners where start time is after end time
  - Real-time updates via PubSub - When banners are created/updated/deleted, all riders on /home see changes instantly
  - Prominent display on rider home page - Banners appear at top.
  - Multiple active banners supported - Shows all currently active banners simultaneously
  - Smart form with split date/time inputs - Follows campaign form pattern for better UX
  - Automatic expiration - Banners disappear when time window ends or when disabled
  - Query optimization - Efficient filtering by time range and enabled status
  - Comprehensive test coverage - Validation, CRUD, active filtering, real-time updates, and display integration

Useful for weather alerts, service updates, announcements, and urgent notices!

### DIFF
--- a/lib/bike_brigade/messaging.ex
+++ b/lib/bike_brigade/messaging.ex
@@ -440,14 +440,14 @@ defmodule BikeBrigade.Messaging do
 
   @doc """
   Returns currently active banners.
-  
+
   A banner is considered active if:
   - It is enabled
   - The current time is between turn_on_at and turn_off_at
   """
   def list_active_banners() do
     now = DateTime.utc_now()
-    
+
     from(b in Banner,
       where: b.enabled == true,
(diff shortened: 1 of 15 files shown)


teesloane:
> What do you think? Or is that too confusing with Messaging and Notifications as different contexts?

a "Notifications" context makes sense! I will move that over.

PR #463, opened 2026-01-20
Feature: Add ability for dispatchers to add signatures to their messages

teesloane:
## Describe your changes

 - Adds optional dispatcher signature feature for outgoing SMS messages
  - Dispatchers can enable "Signature on Messages" in their user settings to auto-append `(sent by: Name)` to messages they send
  - Only affects messages sent by dispatchers through the UI; automated campaign messages are unaffected

My question - do we want to be adding settings on the user like this? Or maybe create a table just for settings, and then assign them in a many-to-many relationship with users? etc

##  Tests

  - New unit tests for maybe_append_signature/2 function
  - Tests for signature_on_messages field in accounts context
  - Manual: Enable signature in user settings, send a message, verify signature appears
  - Manual: Verify campaign messages do not include signatures

## Screenshots

image

### DIFF
--- a/lib/bike_brigade_web/components/conversation_component.ex
+++ b/lib/bike_brigade_web/components/conversation_component.ex
@@ -129,4 +129,22 @@ defmodule BikeBrigadeWeb.Components.ConversationComponent do
     |> List.first()
     |> String.to_atom()
   end
+
+  defp signature_status(assigns) do
+    ~H"""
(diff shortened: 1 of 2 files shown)


mveytsman:
I added a little link in the message footer to be able to change the signature

CleanShot 2026-02-02 at 19 44 37

PR #466, opened 2026-01-26
Add dedicated Tags management page and hide restricted tags from riders

mveytsman:
## Describe your changes

- Add /tags admin page for dispatchers to manage rider tags
- Tags can be marked as restricted, hiding them from riders on their profile
- Restricted tags are shown with a lock icon in the dispatcher view
- Riders can only see and modify non-restricted tags on their profile
- Remove inline tag creation from rider edit form (use Tags page instead)
- Add Tags link to sidebar navigation

CleanShot 2026-01-26 at 00 22 55@2x

Editing a riders' tags
CleanShot 2026-01-26 at 00 23 05@2x

When not a dispatcher you don't see restricted tags
CleanShot 2026-01-26 at 00 23 31@2x

-  [x] I have performed a self-review of my code

### DIFF
--- a/lib/bike_brigade_web/live/rider_live/tags_component.ex
+++ b/lib/bike_brigade_web/live/rider_live/tags_component.ex
@@ -151,7 +151,11 @@ defmodule BikeBrigadeWeb.RiderLive.TagsComponent do
         />
         <.suggested_tag :for={tag <- @suggested_tags} tag={tag} target={@myself} />
       </div>
-      <.link :if={@is_dispatcher} navigate={~p"/tags"} class="text-sm text-gray-500 hover:text-gray-700">
+      <.link
+        :if={@is_dispatcher}
+        navigate={~p"/tags"}
+        class="text-sm text-gray-500 hover:text-gray-700"
+      >
         Manage tags
       </.link>
     </div>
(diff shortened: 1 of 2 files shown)


lib/bike_brigade_web/live/rider_live/tags_component.ex
```diff
@@ -7,80 +7,126 @@ defmodule BikeBrigadeWeb.RiderLive.TagsComponent do
...
+        <span
+          class={[
+            "my-0.5 inline-flex items-center px-2.5 py-1.5 rounded-md text-md font-medium cursor-pointer hover:bg-gray-200",
+            if(tag.restricted, do: "bg-gray-200 text-gray-500", else: "bg-gray-100 text-gray-500")
+          ]}
+          phx-click={JS.push("select", value: %{name: tag.name}, target: @myself)}
+        >
+          <Heroicons.plus mini class="w-4 h-4 mr-1" />
+          {tag.name}
+          <Heroicons.lock_closed :if={tag.restricted} mini class="w-4 h-4 ml-1 text-amber-600" />
```
teesloane: Not required but we could make this a "dumb component" to tidy up a bit.

PR #479, opened 2026-03-17
Replace appearances of "The Island" as neighbourhood

teesloane:
I looked at the GeoJSON for the Waterfront Islands Community Polygon and it maps out to this part of Toronto.

image

so for now I think it does make sense as we discussed in slack to just show "St. Lawrence Market" for addresses within this polygon.


mveytsman:
Closing in favor of #481

PR #484, opened 2026-03-29
Add list_campaigns_ended_between/2 to query campaigns by end time

neal-bpm:
https://github.com/bikebrigade/dispatch/issues/476

## Describe your changes

This function enables finding campaigns whose delivery window ended within a specified time range. It's designed to support a scheduled job that runs every 15 minutes to process recently ended campaigns.

The function:
- Accepts from_datetime and to_datetime as NaiveDateTime
- Uses inclusive bounds (>= and <=) for the time window
- Returns a list of Campaign structs without preloads

Test coverage includes:
- Finding campaigns within the specified window
- Returning empty list when no campaigns match

- [X] I have performed a self-review of my code

### DIFF
--- a/lib/bike_brigade/delivery.ex
+++ b/lib/bike_brigade/delivery.ex
@@ -312,6 +312,24 @@ defmodule BikeBrigade.Delivery do
   end
 
   @doc """
+  Returns campaigns whose delivery window ended within the given time range.
+
+  ## Parameters
+    - `from_datetime` - Start of the time window (inclusive)
+    - `to_datetime` - End of the time window (inclusive)
+  """
+  def list_campaigns_ended_between(from_utc_datetime, to_utc_datetime) do
+    query =
+      from c in Campaign,
+        where:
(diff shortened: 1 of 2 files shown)


lib/bike_brigade/delivery.ex
```diff
@@ -311,6 +311,24 @@ defmodule BikeBrigade.Delivery do
...
   end
 
+  @doc """
+  Returns campaigns whose delivery window ended within the given time range.
+
+  ## Parameters
+    - `from_datetime` - Start of the time window (inclusive)
+    - `to_datetime` - End of the time window (inclusive)
+  """
+  def list_campaigns_ended_between(from_utc_datetime, to_utc_datetime) do
```
mveytsman, line 321: i think the params should be `from_datetime` and `to_datetime` like in the docs above

neal-bpm: I updated the changes in the PR: https://github.com/bikebrigade/dispatch/pull/486

Issue #513, opened 2026-09-01
Bug: becoming a backup rider cancels regular rider assignments


wjyoung65:
**Describe the bug**

Signing up to be a backup rider for a campaign cancels ones deliveries for that campaign.

If signed up to be a backup rider, one cannot sign up for unclaimed deliveries until one has cancelled being a backup rider.

Example: I *thought* I had signed up with SoH to deliver to "E" and "M" last week. They are two of my regulars. I had the time blocked off in my calendar, so I am quite confident I had signed up for them. At the time, I also marked myself down to be a backup rider. Tuesday rolls around, and Rob (SoH) contacts me about two unclaimed deliveries. Turns out they are for "E" and "M". I click on the "Assign me" button, and a popup says I have to unassign myself as a backup rider first. After doing that, I can sign up for E and M. But now I cannot assign myself as a backup rider.

**Steps to Reproduce**

If possible, please provide steps to reproduce the behaviour:

Assign self to a delivery, then assign self as a backup rider.

Or, assign self as a backup rider, then assign self to a delivery.

**Expected behavior**

Describe what you expected to happen.

Being a backup rider should not exclude one from doing deliveries, and vice versa.

**Screenshots**

If applicable, add screenshots of the bug in action!

**Environment (please complete the following information):**

 - OS: [e.g. iOS] MacOS
 - Browser [e.g. chrome, safari]  safari

**Additional context**

This seems to be new behaviour, so it may be related to https://github.com/bikebrigade/dispatch/pull/512 which was recently merged.

wjyoung65:
I asked Claude to investigate, and I will paste its reply below. I don't have a dev/test env setup (yet), but I will try to repro the bug and test the fix.

(start Claude quote)

I dug through the actual code paths involved, and I don't think #512 is the cause — the mechanism is fully explainable by a bug that's been sitting in the codebase since backup riders were added over a year ago, and #512 never touches the code that produces this symptom. Let me show the evidence, then the fix.

Why it's not #512

#512's entire diff is one line, in hacky_assign/1 (the automated bulk-assignment job):
+        where: cr.backup_rider == false,
[...]

PR #514, opened 2026-09-03
Fix backup riders' task assignments resolving to nil (#513)

neal-bpm:
## Describe your changes

When a rider marked themselves as backup, their existing task assignments silently appeared unclaimed on the signup page. The root cause: campaign_riders_and_tasks/1 was reusing all_riders — filtered to backup_rider == false — for the in-memory preload that resolves task.assigned_rider. Backup riders were absent from that list, so the preload resolved their assigned_rider_id to nil despite the database being correct.

  Split into two queries: all_riders stays filtered for the display panel, and a new all_riders_for_resolution — unfiltered, no logistics fields — is used solely for the preload. Backup riders no longer vanish from assignment resolution without affecting what the riders panel shows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed task assignments so riders designated as backups are now correctly displayed when viewing campaign tasks.
  * Preserved the existing display behavior that excludes backup riders from the general rider list.

### DIFF
--- a/lib/bike_brigade/delivery.ex
+++ b/lib/bike_brigade/delivery.ex
@@ -551,8 +551,18 @@ defmodule BikeBrigade.Delivery do
       )
       |> Repo.preload([:location, :total_stats])
 
-    # Does a nested preload to get tasks' assigned riders without doing an extra db query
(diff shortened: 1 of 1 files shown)


mveytsman:
Close for #516

